PR #1, opened 2016-10-25
Better document CALL ... YIELD ... (and WHERE ...)

petraselmer:
Original PR by @boggle at https://github.com/neo4j/neo4j/pull/8237


boggle:
Thanks for picking this up, @petraselmer

PR #5, opened 2016-11-22
Fix typo in replace function description

Gudahtt:
There was a typo in the description for the String function "replace" - it looks like the word "be" was unnecessarily inserted at some point.

Let me know if this was supposed to say "must be" instead of "are" - I wasn't sure which wording was intended, so I went with the simpler option.


petraselmer:
Hi Mark

Thanks for raising this. Please can you make two changes, after which I will merge:

1) It ought to be "must be" instead of "are"
2) Please can you raise the PR against the 3.0 branch, as the error is present thereon in.

Thanks

Gudahtt:
Sure, no problem. I wasn't sure what the policy was on old docs - so I just chose the default branch.

PR #15, opened 2017-01-05
Fix syntax of acos functions

szarnyasg:
This is a fix for a minor issue: the syntax of the `acos` function was incorrectly marked as `abs`.

PR #16, opened 2017-01-10
Fix syntax of acos functions on 3.0

szarnyasg:
Opened a separate PR as discussed in #15


jjaderberg:
Thanks @szarnyasg for finding and fixing this!

PR #21, opened 2017-01-16
Removed dependency on internal kernel IndexDescriptor

fickludd:
Remove dependency on internal kernel api and IndexDescriptor
Fixed broken BlockTypeTest


jjaderberg:
@fickludd Is this an independent change for merge now, or is it staged to go together with one or more product changes?

craigtaverner:
@jjaderberg This change can go in immediately, as it removes a dependency on something that is changing in Neo4j. With the dependency removed, docs should work against both the pre- and post-change versions of Neo4j.

fickludd:
@jjaderberg it's fine to merge

PR #32, opened 2017-02-20
Textual edits to the parameter page

petraselmer:
- Clearer demarcation of where params can and cannot be used
- Deprecation warning

### DIFF
--- a/manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
+++ b/manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
@@ -3,19 +3,32 @@
 
 Cypher supports querying with parameters.
 This means developers don't have to resort to string building to create a query.
-In addition to that, it also makes caching of execution plans much easier for Cypher.
+Additionally, parameters make caching of execution plans much easier for Cypher, thus leading to faster query execution times.
 
-Parameters can be used for literals and expressions in the `WHERE` clause, for the index value in the `START` clause, index queries, and finally for node/relationship ids.
-Parameters can not be used as for property names, relationship types and labels, since these patterns are part of the query structure that is compiled into a query plan.
+Parameters can be used for:
 
-Accepted names for parameters are letters and numbers, and any combination of these.
+* Literals and expressions (such as in the `WHERE` clause)
+* Node and relationship ids
+* For legacy indexes only: index values and queries
+
+Parameters cannot be used for the following constructs, as these form part of the query structure that is compiled into a query plan:
+
+* Property keys; so, `MATCH (n) WHERE n.$param = 'something'` is invalid
+* Relationship types
+* Labels
+
+Parameters may consist of letters and numbers, and any combination of these, but cannot start with a number or a currency symbol.
 
 For details on using parameters via the Neo4j REST API, see <<rest-api-transactional>>.
 // For details on parameters when using the Neo4j embedded Java API, see <<tutorials-cypher-parameters-java>>.
 
-Below follows a comprehensive set of examples of parameter usage.
-The parameters are given as JSON here.
-Exactly how to submit them depends on the driver in use.
+We provide below a comprehensive list of examples of parameter usage.
+In these examples, parameters are given in JSON; the exact manner in which they are to be submitted depends upon the driver being used.
+
+[NOTE]
+====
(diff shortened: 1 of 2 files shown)


petraselmer:
@jjaderberg & @mariascharin please can you take a look at this. I am happy to merge it when you've indicated it is ok.

manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
```diff
@@ -3,19 +3,32 @@
...
 This means developers don't have to resort to string building to create a query.
-In addition to that, it also makes caching of execution plans much easier for Cypher.
+Additionally, parameters make caching of execution plans much easier for Cypher, thus leading to faster query execution times.
 
-Parameters can be used for literals and expressions in the `WHERE` clause, for the index value in the `START` clause, index queries, and finally for node/relationship ids.
-Parameters can not be used as for property names, relationship types and labels, since these patterns are part of the query structure that is compiled into a query plan.
+Parameters can be used for:
 
-Accepted names for parameters are letters and numbers, and any combination of these.
+* Literals and expressions (such as in the `WHERE` clause)
```
jjaderberg: For long form list items please terminate each item as if it were an independent sentence.
In general please use on of two formats for lists:
a)
```
This list:
* Has items with more than just one or two words.
* Capitalizes and terminates each item as if it were an independent sentence.
```
b)
```
This list:
* is terse
* not sentence-styled
```
I'd say the first list fits the first format, the second the second.

manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
```diff
@@ -3,19 +3,32 @@
...
 
-Below follows a comprehensive set of examples of parameter usage.
-The parameters are given as JSON here.
-Exactly how to submit them depends on the driver in use.
+We provide below a comprehensive list of examples of parameter usage.
+In these examples, parameters are given in JSON; the exact manner in which they are to be submitted depends upon the driver being used.
+
+[NOTE]
+====
+It is recommended that the new parameter syntax -- `$param` -- is used, as the old syntax (`{param}`) is deprecated and will be removed altogether in a later release.
```
jjaderberg: Would it make sense to include the good and bad syntax examples the same way, using either `()` or `--`?

jjaderberg (dismissed):
Two minor comments. I can merge when addressed as I see a cross reference in this file that I'd like to update while we're poking here

manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
craigtaverner: I also think the use of `-- ... --` as well as `( ... )` as grouping is a bit odd. I'd prefer something consistent like:

    It is recommended that the new parameter syntax `$param` is used, as the old syntax `{param}` is deprecated and will be removed altogether in a later release.

manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
craigtaverner: I prefer the terse list version here since most of the entries are terse. The existence of a single long entry does not feel like sufficient reason to change all to long form.

craigtaverner: Is there value in adding an in-line example here or rely on people to look below? I say this only because the first thing I thought when reading `such as in the WHERE clause` was _show me an example_.

craigtaverner:
What about parameters for procedures? Is that handled somewhere else?

manual/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
petraselmer: @craigtaverner I agree. I'm making both lists terse

petraselmer: @craigtaverner I'm also removing the `WHERE` snippet: it adds nothing

petraselmer:
@craigtaverner there is now an example illustrating the use of parameters in a procedure call

jjaderberg (approved):
LGTM, and good call adding labels!

PR #34, opened 2017-02-22
Get (back) the correct default and unit for DURATION settings

jjaderberg:
Based on neo4j/neo4j#8875

### DIFF
--- a/manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
+++ b/manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
@@ -20,17 +20,18 @@
 package org.neo4j.doc;
 
 import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 import org.neo4j.graphdb.config.Setting;
 import org.neo4j.graphdb.factory.Description;
+import org.neo4j.helpers.TimeUtil;
 import org.neo4j.kernel.configuration.Group;
 import org.neo4j.kernel.configuration.Internal;
 import org.neo4j.kernel.configuration.Obsoleted;
+import org.neo4j.kernel.configuration.Settings;
 
 /**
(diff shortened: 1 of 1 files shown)


manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
```diff
@@ -20,17 +20,18 @@
 package org.neo4j.doc;
 
 import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
spacecowboy: glob-import here (change your intellij settings to avoid glob-imports)

manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
```diff
@@ -76,6 +78,11 @@ public static SettingsDescription describe( Class<?> settingClass )
                 {
                     Object rawDefault = setting.apply( from -> null );
                     defaultValue = rawDefault != null ? rawDefault.toString() : null;
+                    // Kludge--because for DURATION settings, the internal representation (MS) is leaked as the default value
+                    Optional<? extends Function<String, ?>> parser = setting.getParser();
+                    if (null != defaultValue && parser.isPresent() && Settings.DURATION.equals(parser.get())) {
+                        defaultValue = Long.toString(TimeUtil.DEFAULT_TIME_UNIT.convert(Long.valueOf(defaultValue), TimeUnit.MILLISECONDS));
```
spacecowboy, line 87: I think we should add the unit to the string also, for extra clarity

spacecowboy (changes requested):

manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
jjaderberg: sure

manual/doctools/src/main/java/org/neo4j/doc/SettingsDescription.java
jjaderberg: there's no immediate way to get the unit from TimeUtil. it's a good-to-have so let's defer it to the new tool in 3.2

PR #37, opened 2017-02-25
Replace property-container to expression that returns a node/relation…

### LINKED ISSUE #36: property-container input is deprecated
According to the documentation, the [`keys()`](https://neo4j.com/docs/developer-manual/current/cypher/functions/list/#functions-keys) and the [`id()`](https://neo4j.com/docs/developer-manual/current/cypher/functions/scalar/#functions-id) functions expect a `property-container` as their inputs, while the [`properties()`](https://neo4j.com/docs/developer-manual/current/cypher/functions/scalar/#functions-properties) function expects an `expression`, which is _"an expression that returns a node, a relationship, or a map"_.

Of these, only the latter is precise - both `keys()` and `properties()` accept expressions as their inputs (tested with 3.0.7 and 3.1.1).

Using this dataset,
```
CREATE ()-[r:REL {name: "x", value: 1}]->()
```
running this query,
```
MATCH ()-[r*]->()
RETURN properties(r[0]), keys(r[0]), id(r[0])
```
results in:
```
╒════════════════════════╤════════════════╤══════════╕
│"properties(r[0])"      │"keys(r[0])"    │"id(r[0])"│
╞════════════════════════╪════════════════╪══════════╡
│{"name":"x","value":"1"}│["name","value"]│"85"      │
└────────────────────────┴────────────────┴──────────┘
```

I think the `property-container` input (used in the documentation) is no longer valid - all of these functions allow the user to specify an expression returning the appropriate elements (nodes/relationships and for the `property()` function, maps).

szarnyasg:
…ship(/map)

Fixes #36.

### DIFF
--- a/manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/FunctionsTest.scala
+++ b/manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/FunctionsTest.scala
@@ -188,8 +188,8 @@ class FunctionsTest extends DocumentingTestBase {
   @Test def keys() {
     testThis(
       title = "keys()",
-      syntax = "keys(  property-container )",
-      arguments = List("property-container" -> "A node, a relationship, or a literal map."),
+      syntax = "keys(  expression )",
+      arguments = List("expression" -> "An expression that returns a node, a relationship, or a map"),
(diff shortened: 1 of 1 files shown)


petraselmer:
Thanks @szarnyasg - looks good!

PR #38, opened 2017-03-01
Remove pin count metric configuration note

MishaDemianenko:
Remove additional configuration for page cache pin tracing since
now we always count page cache pins.


jjaderberg (approved):
LGTM

PR #39, opened 2017-03-06
Fix bold cell text in config settings

mariascharin:


mariascharin:
Re-submit against 3.0.

PR #44, opened 2017-03-10
Remove tests that are not part of the documentation

chrisvest:
… and were breaking due to recent changes in the product.

The build failed locally, though. I don't know why, because the failure made no sense to me.

Please take a look, @jjaderberg


jjaderberg:
Thanks @chrisvest there are multiple breaking changes at the moment so it's hard to prove fixes. This looks sound but it I think it should go into 3.0 so I cherry picked and merged it; ready, fire, aim.

PR #49, opened 2017-03-20
Draw out new List sub-sections on List landing page

petraselmer:
This must be fwd-merged. 

No chunking required.


petraselmer:
@jjaderberg luckily this turned out not to require chunking, only the drawing out of pattern comprehensions and map projections (introduced in 3.1). So, I think this is a straightforward fwd-merge.

PR #63, opened 2017-04-12
Deprecated ArticleTest

petraselmer:
To be merged in conjunction with https://github.com/neo-technology/neo4j-manual-modeling/pull/252


petraselmer:
@jjaderberg I will raise a sep. PR for 3.0 only, which will then need to be null fwd-merged. This will make it easier (as one file was split into two in 3.1)

jjaderberg (approved):

PR #91, opened 2017-06-19
Update year in NOTICE files

Mats-SX:
Also does a few other fixes, which were required for maven to pass compilation.


Mats-SX:
@jjaderberg What do we do with this one? Close?

jjaderberg:
@Mats-SX Sorry, yes I'll close it. I cherry-picked one commit and made equivalent changes to the other from 3.2 on.

PR #113, opened 2017-08-16
Updated docs after refactoring

pontusmelke:
A lot of the cypher internals are being shuffled around, this commit
updates the cypher docs to cope with that.

This PR should be merged at the same time as https://github.com/neo4j/neo4j/pull/9780 to avoid a lot of red builds.


jjaderberg (approved):
LGTM

PR #114, opened 2017-08-18
Disable all license checking

Mats-SX:
These maven modules are never published anywhere, and do not require
adherence to OSS license terms.


Mats-SX:
Superseded by #115

PR #115, opened 2017-08-18
Disable artifact license checking

Mats-SX:
These maven modules are never published anywhere, and do not require adherence to OSS license terms.


jjaderberg (approved):
👍

PR #116, opened 2017-08-21
3.1 function tidyups

petraselmer:


petraselmer:
Added 'DO NOT MERGE' label pending review of correctness by @craigtaverner 

@mariascharin please feel free to review as well, in the meantime, especially wrt formatting and consistency. It's a large PR and I am sure errors are lurking in there somewhere.

sarmbruster (approved):
looks fine.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
       }
     }
     section("collect()", "functions-collect") {
       p(
-        """`collect()` collects all the values into a list.
-          |It will ignore `null` values.""".stripMargin)
-      function("collect(expression)", ("expression", "An expression."))
+        """`collect()` returns a list containing the values returned by an expression.
+          |Using this function aggregates data by amalgamating multiple records or values into a single list.""".stripMargin)
+      function("collect(expression)", "A list containing heterogeneous elements; the types of the elements are determined by the values returned by`expression`.", ("expression", "An expression returning a set of values."))
```
mariascharin: Minimal typo, should be: ...returned by `expression`.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
 |<<functions-degrees, degrees()>>               | Trigonometric     | Converts radians to degrees.
-|<<functions-distance,distance()>>              | Spatial           | Returns a floating point number representing the geodesic distance between any two given points.
+|<<functions-distance,distance()>>              | Spatial           | Returns a floating point number representing the geodesic distance between any two given points in the same CRS.
 |<<functions-e, e()>>                           | Logarithmic       | Returns the base of the natural logarithm, `e`.
-|<<functions-endnode,endNode()>>                | Scalar            | Returns the last node of a relationship.
-|<<functions-exists,exists()>>                  | Predicate         | Returns true if a match for the pattern exists in the graph, or the property exists in the node, relationship or map.
+|<<functions-endnode,endNode()>>                | Scalar            | Returns the end node of a relationship.
+|<<functions-exists,exists()>>                  | Predicate         | Returns true if a match for a given pattern exists in the graph, or if a specified property exists in the node, relationship or map.
```
mariascharin: We are not consistent in the formatting of boolean values in text. In most of the Ops manual we format it as `true`/`false`. We should decide on a format and document it in the styleguide.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-abs, abs()>>                       |  Numeric          | Returns the absolute value of a given number.
+|<<functions-acos, acos()>>                     | Trigonometric     | Returns the arccosine of a number in radians.
+|<<functions-all,all()>>                        | Predicate         | Tests whether a predicate holds for all elements in a given list.
+|<<functions-any,any()>>                        | Predicate         | Tests whether a predicate holds for at least one element in a given list.
+|<<functions-asin, asin()>>                     | Trigonometric     | Returns the arcsine of a number in radians.
+|<<functions-atan, atan()>>                     | Trigonometric     | Returns the arctangent of a number in radians.
+|<<functions-atan2, atan2()>>                   | Trigonometric     | Returns the arctangent2 of a set of coordinates in radians.
+|<<functions-avg,avg()>>                        | Aggregating       | Returns the average of a set of values.
+|<<functions-ceil, ceil()>>                     | Numeric           | Returns the smallest floating point number greater than or equal to a given number, and which is equal to an integer.
```
mariascharin: This is slightly confusing/cumbersome to read (several places; applies also to `floor()`). Can it be expressed in a simpler way?

Seems like the equivalent functions in SQL actually return integers rather than floats:
(... some googling...)
The CEILING function returns the smallest integer greater than or equal to the specified numeric expression. 
The FLOOR function returns the largest integer less than or equal to the specified numeric expression.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
-|<<functions-round, round()>>                   | Numeric           | Returns the numerical expression, rounded to the nearest integer.
-|<<functions-rtrim,rTrim()>>                    | String            | Returns the original string with whitespace removed from the right side.
-|<<functions-sign, sign()>>                     | Numeric           | Returns the signum of a number -- zero if the expression is zero, `-1` for any negative number, and `1` for any positive number.
-|<<functions-sin, sin()>>                       | Trigonometric     | Returns the sine of the expression.
-|<<functions-single,single()>>                  | Predicate         | Returns true if the predicate holds for exactly one of the elements in the list.
+|<<functions-rand, rand()>>                     | Numeric           | Returns a random floating point number in the range from 0 (inclusive) to 1 (exclusive); i.e. [0, 1).
+|<<functions-range,range()>>                    | List              | Returns a list comprising all integral values within a range bounded by a start and end value, where the difference between any two consecutive values is constant.
```
mariascharin: `integral` seems to be used for "integer values", but this is not the way I understand the mathematic term. Googling confirms this. Pls verify.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -47,21 +47,21 @@ These functions return a single value.
...
 | <<functions-head,head()>>                                           | Returns the first element in a list.
-| <<functions-id,id()>>                                               | Returns the id of the relationship or node.
+| <<functions-id,id()>>                                               | Returns the id of a relationship or node.
 | <<functions-last,last()>>                                           | Returns the last element in a list.
 | <<functions-length,length()>>                                       | Returns the length of a path.
-| <<functions-properties,properties()>>                               | If the argument is a node or a relationship, the returned map is a map of its properties.
+| <<functions-properties,properties()>>                               | Returns a map containing all the properties of a node or relationship.
 | <<functions-size,size()>>                                           | Returns the number of items in a list.
 | <<functions-size-of-pattern-expression,Size of pattern expression>> | Returns the number of sub-graphs matching the pattern expression.
```
mariascharin: For uniformity of the table this might instead be written as "size() applied to pattern expression"
 (just an idea; not necessary)

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -47,21 +47,21 @@ These functions return a single value.
...
 | <<functions-size-of-string,Size of string>>                         | Returns the size of a string.
```
mariascharin: For uniformity of the table this might instead be written as "size() applied to string"
 (just an idea; not necessary)

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -47,21 +47,21 @@ These functions return a single value.
...
+| <<functions-coalesce,coalesce()>>                                   | Returns the first non-`null` value in a given list of expressions.
+| <<functions-endnode,endNode()>>                                     | Returns the end node of a relationship.
 | <<functions-head,head()>>                                           | Returns the first element in a list.
-| <<functions-id,id()>>                                               | Returns the id of the relationship or node.
+| <<functions-id,id()>>                                               | Returns the id of a relationship or node.
 | <<functions-last,last()>>                                           | Returns the last element in a list.
 | <<functions-length,length()>>                                       | Returns the length of a path.
-| <<functions-properties,properties()>>                               | If the argument is a node or a relationship, the returned map is a map of its properties.
+| <<functions-properties,properties()>>                               | Returns a map containing all the properties of a node or relationship.
```
mariascharin, line 56: It is probably implicit that "a map containing all the properties of a node or relationship" will contain both key and value for each property but it may not be obvious for e.g. a SQL person. Is it possible to spell it out?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -185,36 +194,39 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
     section("percentileCont()", "functions-percentilecont") {
       p(
-        """`percentileCont()` calculates the percentile of a given value over a group, with a percentile from 0.0 to 1.0.
-          |It uses a linear interpolation method, calculating a weighted average between two values, if the desired percentile lies between them.
+        """`percentileCont()` returns the percentile of a given value over a group, with a percentile from 0.0 to 1.0.
+          |It uses a linear interpolation method, calculating a weighted average between two values if the desired percentile lies between them.
           |For nearest values using a rounding method, see `percentileDisc`.""")
-      function("percentileCont(expression, percentile)", ("expression", "A numeric expression."), ("percentile", "A numeric value between 0.0 and 1.0"))
+      function("percentileCont(expression, percentile)", "A Float.", ("expression", "A numeric expression."), ("percentile", "A numeric value between 0.0 and 1.0"))
+      considerations("Any `null` values are excluded from the calculation.", "`percentileCont(null, <percentile>)` returns `null`.")
```
mariascharin: It seems like we are leaving the angular brackets out in the rest of the text, so should probably do here as well (two places). Also applies to <some-path> further down.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/MathematicalTrigonometricFunctionsTest.scala
```diff
@@ -45,80 +45,87 @@ class MathematicalTrigonometricFunctionsTest extends DocumentingTest {
...
+        p("The arccosine of `0.5` is returned.")
         resultTable()
       }
     }
     section("asin()", "functions-asin") {
-      p("`asin()` returns the arcsine of the expression, in radians.")
-      function("asin(expression)", ("expression", "A numeric expression that represents the angle in radians."))
+      p("`asin()` returns the arcsine of a number in radians.")
+      function("asin(expression)", "A Float.", ("expression", "A numeric expression that represents the angle in radians."))
+      considerations("`asin(null)` returns `null`.", "If `expression` < -1 or `expression` > 1, `asin(expression)` returns `null`.")
```
mariascharin: Slightly difficult to read. Can it be written with parenthesis, i.e.
If (expression < -1) or (expression > 1) then acos(expression) returns null (just an idea; several places).

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/MathematicalTrigonometricFunctionsTest.scala
```diff
@@ -45,80 +45,87 @@ class MathematicalTrigonometricFunctionsTest extends DocumentingTest {
...
     }
     section("haversin()", "functions-haversin") {
-      p("`haversin()` returns half the versine of the expression.")
-      function("haversin(expression)", ("expression", "A numeric expression that represents the angle in radians."))
+      p("`haversin()` returns half the versine of a number.")
+      function("haversin(expression)", "A Float.", ("expression", "A numeric expression that represents the angle in radians."))
+      considerations("`haversin(null)` returns `null`.")
       query("RETURN haversin(0.5)", ResultAssertions((r) => {
         r.toList.head("haversin(0.5)") should equal(0.06120871905481362)
       })) {
```
mariascharin, line 131: In the comment to the haversin() example further down (not "commentable" b/c not in PR scope), Berlin and San Mateo is formatted as *'Berlin'* and *'San Mateo'* which is according to the style in the rest of the chapter.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -174,8 +182,9 @@ class AggregatingFunctionsTest extends DocumentingTest {
       }
     }
     section("min()", "functions-min") {
-      p("`min()` takes a numeric property as input, and returns the lowest value in that column.")
-      function("min(expression)", ("expression", "A numeric expression."))
+      p("`min()` returns the minimum value in a set of values.")
+      function("min(expression)", "Either an Integer or a Float or a String, depending on the values returned by `expression`.", ("expression", "An expression returning a set of numeric or string values."))
+      considerations("Any `null` values are excluded from the calculation.", "`min(null)` returns `null`.")
```
mariascharin, line 188: The example is not really helpful since you can't see the data model or sample data! This is applicable to a fair number of the examples in this chapter (e.g. 3.4.2.1 (coalesce()) and 3.4.2.10 (size()). In some other examples, e.g. the one in 3.4.2.7 properties(), the data is created _within_ the example so the result is illustrative. Similarly in 3.4.2.15 toInteger(), the example is very clear and easy to understand.

mariascharin (commented):
Comments on many different levels. Let's discuss.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
-|<<functions-ltrim,lTrim()>>                    | String            | Returns the original string with whitespace removed from the left side.
-|<<functions-max,max()>>                        | Aggregating       | Returns the highest value in a numeric column.
-|<<functions-min,min()>>                        | Aggregating       | Returns the lowest value in a numeric column.
-|<<functions-nodes,nodes()>>                    | List              | Returns all nodes in a path.
-|<<functions-none,none()>>                      | Predicate         | Returns true if the predicate holds for no element in the list.
+|<<functions-log, log()>>                       | Logarithmic       | Returns the natural logarithm of a number.
+|<<functions-log10, log10()>>                   | Logarithmic       | Returns the common logarithm (base 10) of a number.
+|<<functions-ltrim,lTrim()>>                    | String            | Returns the original string with leading whitespace removed.
+|<<functions-max,max()>>                        | Aggregating       | Returns the maximum value in a set of values.
```
craigtaverner: I thought max was an aggregating function, but the new description seems to imply it takes a single list argument and returns the max element, which is not what it does. It should take a column identifier (or variable) and aggregated the column (reducing cardinality) to return a single value (max value), based on any other grouping conditions.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
 |<<functions-length,length()>>                  | Scalar            | Returns the length of a path.
-|<<functions-log, log()>>                       | Logarithmic       | Returns the natural logarithm of the expression.
-|<<functions-log10, log10()>>                   | Logarithmic       | Returns the common logarithm (base 10) of the expression.
-|<<functions-ltrim,lTrim()>>                    | String            | Returns the original string with whitespace removed from the left side.
-|<<functions-max,max()>>                        | Aggregating       | Returns the highest value in a numeric column.
-|<<functions-min,min()>>                        | Aggregating       | Returns the lowest value in a numeric column.
-|<<functions-nodes,nodes()>>                    | List              | Returns all nodes in a path.
-|<<functions-none,none()>>                      | Predicate         | Returns true if the predicate holds for no element in the list.
+|<<functions-log, log()>>                       | Logarithmic       | Returns the natural logarithm of a number.
```
craigtaverner: Replacing 'the expression' with 'a number' makes it almost seem like you need to pass a literal here. But I guess 'an expression that evaluates to a number' is too ugly.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-log10, log10()>>                   | Logarithmic       | Returns the common logarithm (base 10) of a number.
+|<<functions-ltrim,lTrim()>>                    | String            | Returns the original string with leading whitespace removed.
+|<<functions-max,max()>>                        | Aggregating       | Returns the maximum value in a set of values.
+|<<functions-min,min()>>                        | Aggregating       | Returns the minimum value in a set of values.
+|<<functions-nodes,nodes()>>                    | List              | Returns a list containing all the nodes in a path.
+|<<functions-none,none()>>                      | Predicate         | Returns true if the predicate holds for no element in a given list.
```
craigtaverner: Prefer the original. Changing 'the list', but not also changing 'the predicate' makes it inconsistent. This function is very complex, and not easy to describe with a single line, so I prefer the terseness of the original.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-exists,exists()>>                  | Predicate         | Returns true if a match for a given pattern exists in the graph, or if a specified property exists in the node, relationship or map.
 |<<functions-exp, exp()>>                       | Logarithmic       | Returns `e^n`, where `e` is the base of the natural logarithm, and `n` is the value of the argument expression.
-|<<functions-extract,extract()>>                | List              | Returns a single property, or the value of a function from a list of nodes or relationships.
-|<<functions-filter,filter()>>                  | List              | Returns all the elements in a list complying with a predicate.
-|<<functions-floor, floor()>>                   | Numeric           | Returns the greatest integer less than or equal to the expression.
-|<<functions-haversin, haversin()>>             | Trigonometric     | Returns half the versine of the expression.
+|<<functions-extract,extract()>>                | List              | Returns a list `l~result~` containing the values resulting from an expression which has been applied to each element in a list `list`.
```
craigtaverner, line 124: What does the `l~result~` mean (or render to)?

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
-|<<functions-reverse,reverse()>>                | String            | Returns the original string reversed.
-|<<functions-right,right()>>                    | String            | Returns a string containing the right n characters of the original string.
-|<<functions-round, round()>>                   | Numeric           | Returns the numerical expression, rounded to the nearest integer.
-|<<functions-rtrim,rTrim()>>                    | String            | Returns the original string with whitespace removed from the right side.
-|<<functions-sign, sign()>>                     | Numeric           | Returns the signum of a number -- zero if the expression is zero, `-1` for any negative number, and `1` for any positive number.
-|<<functions-sin, sin()>>                       | Trigonometric     | Returns the sine of the expression.
-|<<functions-single,single()>>                  | Predicate         | Returns true if the predicate holds for exactly one of the elements in the list.
+|<<functions-rand, rand()>>                     | Numeric           | Returns a random floating point number in the range from 0 (inclusive) to 1 (exclusive); i.e. [0, 1).
```
craigtaverner: Is removal of the backticks desirable?

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
craigtaverner: The new version is obviously much more explicit and correct, but do we really need to be that explicit? It does make for a very long description. Are there any length policies for this table of descriptions? There are other places where I think detailed descriptions would bring more value (like the all(), none(), any() functions for example).

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-range,range()>>                    | List              | Returns a list comprising all integral values within a range bounded by a start and end value, where the difference between any two consecutive values is constant.
+|<<functions-reduce,reduce()>>                  | List              | Returns the value resulting from the application of an expression on each successive element in a list in conjunction with the result of the computation thus far.
```
craigtaverner: I actually found the original a bit easier to read and understand.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-relationships,relationships()/rels()>>    | List              | Returns a list containing all the relationships in a path.
+|<<functions-replace,replace()>>                | String            | Returns a string in which all occurrences of a specified string in the original string have been replaced by another string.
+|<<functions-reverse,reverse()>>                | String            | Returns a string in which all characters in the original string have been reversed.
```
craigtaverner: Except you do not reverse the characters, but the order of the characters, or in other words reverse the 'string', which the original description said better, I thought.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-right,right()>>                    | String            | Returns a string containing the specified number of rightmost characters of the original string.
+|<<functions-round, round()>>                   | Numeric           | Returns the value of a given number rounded to the nearest integer.
```
craigtaverner: This is a bit like a comment I made above for several edits where the new version almost implies the function must take a literal numerical value, while the original text emphasises that an expression can be passed in.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
-|<<functions-startnode,startNode()>>            | Scalar            | Returns the first node of a relationship.
+|<<functions-startnode,startNode()>>            | Scalar            | Returns the start node of a relationship.
 |<<functions-stdev,stDev()>>                    | Aggregating       | Returns the standard deviation for a given value over a group for a sample of a population.
 |<<functions-stdevp,stDevP()>>                  | Aggregating       | Returns the standard deviation for a given value over a group for an entire population.
-|<<functions-substring,substring()>>            | String            | Returns a substring of the original, with a 0-based index start and length.
-|<<functions-sum,sum()>>                        | Aggregating       | Returns the sum of a numeric column.
-|<<functions-tail,tail()>>                      | List              | Returns all but the first element in a list.
-|<<functions-tan, tan()>>                       | Trigonometric     | Returns the tangent of the expression.
+|<<functions-substring,substring()>>            | String            | Returns a substring of the original string, beginning  with a 0-based index start and length..
```
craigtaverner: Extra '.', and I think the extra words don't add any clarity.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-sum,sum()>>                        | Aggregating       | Returns the sum of a set of non-`null` numeric values.
+|<<functions-tail,tail()>>                      | List              | Returns a list `l~result~` containing all the elements, excluding the first one, from a list `list`.
```
craigtaverner: Much more explicit, but does it improve clarity? I'm not convinced.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -101,83 +101,83 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-tan, tan()>>                       | Trigonometric     | Returns the tangent of a number.
 |<<functions-timestamp,timestamp()>>            | Scalar            | Returns the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC.
-|<<functions-toboolean,toBoolean()>>                | Scalar            | Converts the argument to a boolean and returns the result.
-|<<functions-tofloat,toFloat()>>                | Scalar            | Converts the argument to a float and returns the result.
-|<<functions-tointeger,toInteger()>>                    | Scalar            | Converts the argument to an integer and returns the result.
+|<<functions-toboolean,toBoolean()>>                | Scalar            | Casts a string value to a boolean value.
```
craigtaverner: This change makes it more concise! Great to drop the pointless 'returns the result'

craigtaverner: Although I think your use of 'cast' here and below is perhaps less ideal than 'converts'. I seem to think of casting as something that can be done without 'coercion' or 'conversion', but in these functions, we are doing something more complex, and so 'conversion' or at least 'coercion' would fit better.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -47,21 +47,21 @@ These functions return a single value.
...
 | <<functions-size-of-pattern-expression,Size of pattern expression>> | Returns the number of sub-graphs matching the pattern expression.
 | <<functions-size-of-string,Size of string>>                         | Returns the size of a string.
-| <<functions-startnode,startNode()>>                                 | Returns the first node of a relationship.
+| <<functions-startnode,startNode()>>                                 | Returns the start node of a relationship.
 | <<functions-timestamp,timestamp()>>                                 | Returns the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC.
-| <<functions-toboolean,toBoolean()>>                                 | Converts the argument to a boolean and returns the result.
-| <<functions-tofloat,toFloat()>>                                     | Converts the argument to a float and returns the result.
-| <<functions-tointeger,toInteger()>>                                 | Converts the argument to an integer and returns the result.
+| <<functions-toboolean,toBoolean()>>                                 | Casts a string value to a boolean value.
```
craigtaverner: See my comments above, I think 'cast' implies a very non-invasive change, but 'converts' or at least 'coerces' is more appropriate in these cases.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -74,16 +74,16 @@ These functions take multiple values as arguments, and calculate and return an a
 [options="header"]
 |===
 |Function                                       | Description
-|<<functions-avg,avg()>>                        | Returns the average of a numeric column.
-|<<functions-collect,collect()>>                | Returns a list containing all collected values.
-|<<functions-count,count()>>                    | Returns the number of rows.
-|<<functions-max,max()>>                        | Returns the highest value in a numeric column.
-|<<functions-min,min()>>                        | Returns the lowest value in a numeric column.
+|<<functions-avg,avg()>>                        | Returns the average of a set of numeric values.
```
craigtaverner: We need to be careful not to imply that these functions take lists and return a single value. They take column (or variable) expressions and use that to aggregate many records down to a single record (based on grouping expressions). The changes you have made here, with the use of the word 'set' seem to imply a real function, not an aggregating expression.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -74,16 +74,16 @@ These functions take multiple values as arguments, and calculate and return an a
...
+|<<functions-collect,collect()>>                | Returns a list containing the values returned by a given expression.
+|<<functions-count,count()>>                    | Returns the number of values or rows.
+|<<functions-max,max()>>                        | Returns the maximum value in a set of values.
+|<<functions-min,min()>>                        | Returns the minimum value in a set of values.
```
craigtaverner: I think we really need to return to the use of the word column. Or some equivalent way of highlighting that these are not functions, but aggregating expressions.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -74,16 +74,16 @@ These functions take multiple values as arguments, and calculate and return an a
...
 |<<functions-percentilecont,percentileCont()>>  | Returns the percentile of a given value over a group using linear interpolation.
 |<<functions-percentiledisc,percentileDisc()>>  | Returns the nearest value to a given percentile over a group using a rounding method.
 |<<functions-stdev,stDev()>>                    | Returns the standard deviation for a given value over a group for a sample of a population.
 |<<functions-stdevp,stDevP()>>                  | Returns the standard deviation for a given value over a group for an entire population.
-|<<functions-sum,sum()>>                        | Returns the sum of a numeric column.
+|<<functions-sum,sum()>>                        | Returns the sum of a set of non-`null` numeric values.
```
craigtaverner: Surely 'non-null' is relevant to more than just this expression? It could be true of all aggregating expressions, right?

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -151,18 +151,18 @@ All trigonometric functions operate on radians, unless otherwise specified.
 [options="header"]
 |===
 | Function                           | Description
-| <<functions-acos, acos()>>         | Returns the arccosine of the expression.
-| <<functions-asin, asin()>>         | Returns the arcsine of the expression.
-| <<functions-atan, atan()>>         | Returns the arctangent of the expression.
-| <<functions-atan2, atan2()>>       | Returns the arctangent2 of a set of coordinates.
-| <<functions-cos, cos()>>           | Returns the cosine of the expression.
-| <<functions-cot, cot()>>           | Returns the cotangent of the expression.
+| <<functions-acos, acos()>>         | Returns the arccosine of a number in radians.
```
craigtaverner, line 154: I think I've commented elsewhere about my concerns regarding the removal of the term 'expression'. These comments apply here to all these changes too.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
-      function("collect(expression)", ("expression", "An expression."))
+        """`collect()` returns a list containing the values returned by an expression.
+          |Using this function aggregates data by amalgamating multiple records or values into a single list.""".stripMargin)
+      function("collect(expression)", "A list containing heterogeneous elements; the types of the elements are determined by the values returned by`expression`.", ("expression", "An expression returning a set of values."))
+      considerations("Any `null` values are ignored and will not be added to the list.", "`collect(null)` returns an empty list.")
       query("MATCH (n:Person) RETURN collect(n.age)", ResultAssertions((r) => {
         r.toList.head("collect(n.age)") should equal(Seq(13, 33, 44))
       })) {
-        p("Returns a single row, with all the values collected.")
+        p("All the values are collected and returned in a single list.")
```
craigtaverner, line 108: Removal of the phrase 'single row' moves away from the important point that aggregating functions take many rows and reduce the cardinality of the result set. The new wording could equally well apply to the case where we have a collection in one row, and end with a single list of the collection, still in one row.

craigtaverner: What I'm trying to say here, and above, is the increased uniformity with which we use the term 'set' or 'list' feels like it is hiding the critical difference between a list existing entirely within one row, and a column, or set of results across many rows. I think this difference needs to be emphasized, not hidden.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
         })) {
-          p("This returns the start node and the count of related nodes.")
+          p("The labels and `age` property of the start node `n` and the number of nodes related to `n` are returned.")
           resultTable()
         }
       }
-      section("Group count relationship types") {
-        p("To count the groups of relationship types, return the types and count them with `count(*)`.")
+      section("Using `count(*)` to group and count relationship types") {
+        p("`count(*)` can be used to group relationship types and return the number of each of these.")
```
craigtaverner: I think if you remove the suffix ' of these', it will read better.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
         })) {
-          p("The relationship types and their group count is returned.")
+          p("The relationship types and their group count are returned.")
           resultTable()
         }
       }
-      section("Count entities") {
-        p("Instead of counting the number of results with `count(*)`, it might be more expressive to include the name of the variable you care about.")
+      section("Using `count(expression)` to return the number of values") {
+        p("Instead of simply returning the number of rows with `count(*)`, it may be more useful to return the actual number of values returned by an expression.")
```
craigtaverner, line 140: Changing 'variable' to 'expression' is not wrong, but I think moves away from the specific example being described.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
           })) {
+          p("Both `B` and `C` know `D` and thus `D` will get counted twice when not using `DISTINCT`.")
           resultTable()
         }
       }
     }
     section("max()", "functions-max") {
-      p("`max()` takes a numeric property as input, and returns the highest value in that column.")
-      function("max(expression)", ("expression", "A numeric expression."))
+      p("`max()` returns the maximum value in a set of values.")
```
craigtaverner, line 175: This sounds wrong. We need to remember that these aggregating functions do not work on sets or lists but on columns.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -84,88 +84,96 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+      function("max(expression)", "Either an Integer or a Float or a String, depending on the values returned by `expression`.", ("expression", "An expression returning a set of numeric or string values."))
```
craigtaverner, line 176: Again, this sounds wrong. It sounds like I could call `max([1,2,3])`, because the expression `[1,2,3]` does indeed _return  a set of numeric or string values_. But that is not how aggregating functions work. You should call `max(n.age)`, where the expression acts on a column and aggregates over all rows.

craigtaverner (dismissed):

szarnyasg:
@petraselmer some functions have multiple names, and it's often not immediately obvious which one is standard openCypher and which one will be deprecated.

For example, based on the [openCypher standardisation scope](https://github.com/opencypher/openCypher/blob/141869b501eec80e433432f702f5fb40b1ee7b02/docs/standardisation-scope.adoc) I can tell that `lower()` is deprecated in favor of `toLower()`, but I had to dig [deeper](https://github.com/opencypher/openCypher/blob/141869b501eec80e433432f702f5fb40b1ee7b02/cip/2.testable/CIP2016-07-07-Type-conversion-functions.adoc) to find notes on `toInt()` being deprecated. Maybe it would be beneficial to add a section on this to the docs?

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
petraselmer: Revert to orig in this instance

petraselmer:
@szarnyasg This is something many people have raised, and I am pleased to say that there will be a 'Deprecations' page with all this information - on my list to do within the next 4 weeks.

petraselmer:
@craigtaverner & @mariascharin I have hopefully addressed - as per our discussions - all your feedback. Please check and indicate whether this is ready to be merged.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -100,84 +100,84 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
+|<<functions-atan, atan()>>                     | Trigonometric     | Returns the arctangent of a number in radians.
+|<<functions-atan2, atan2()>>                   | Trigonometric     | Returns the arctangent2 of a set of coordinates in radians.
+|<<functions-avg,avg()>>                        | Aggregating       | Returns the average of a set of values.
+|<<functions-ceil, ceil()>>                     | Numeric           | Returns the smallest floating point number greater than or equal to a number.
+|<<functions-coalesce,coalesce()>>              | Scalar            | Returns the first non-`null` value in a list of expressions.
+|<<functions-collect,collect()>>                | Aggregating       | Returns a list containing the values returned by an expression.
+|<<functions-cos, cos()>>                       | Trigonometric     | Returns the cosine of a number.
+|<<functions-cot, cot()>>                       | Trigonometric     | Returns the cotangent of a number.
+|<<functions-count,count()>>                    | Aggregating       | Returns the number of values or rows.
```
alexaverbuch, line 117: why `values` OR `rows`? 
also, do we want to mention `null` here too?
Something like `"Returns the number of rows that contain a non-null value."`?

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -100,84 +100,84 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
 |<<functions-degrees, degrees()>>               | Trigonometric     | Converts radians to degrees.
-|<<functions-distance,distance()>>              | Spatial           | Returns a floating point number representing the geodesic distance between any two given points.
+|<<functions-distance,distance()>>              | Spatial           | Returns a floating point number representing the geodesic distance between any two points in the same CRS.
 |<<functions-e, e()>>                           | Logarithmic       | Returns the base of the natural logarithm, `e`.
-|<<functions-endnode,endNode()>>                | Scalar            | Returns the last node of a relationship.
-|<<functions-exists,exists()>>                  | Predicate         | Returns true if a match for the pattern exists in the graph, or the property exists in the node, relationship or map.
+|<<functions-endnode,endNode()>>                | Scalar            | Returns the end node of a relationship.
+|<<functions-exists,exists()>>                  | Predicate         | Returns true if a match for a pattern exists in the graph, or if a specified property exists in the node, relationship or map.
```
alexaverbuch: this feels a little inconsistent
`A pattern` vs `A SPECIFIED property`
`A pattern` & `A specified property` vs `THE node`

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -100,84 +100,84 @@ This section comprises a glossary of all the keywords -- grouped by category and
...
-|<<functions-round, round()>>                   | Numeric           | Returns the numerical expression, rounded to the nearest integer.
-|<<functions-rtrim,rTrim()>>                    | String            | Returns the original string with whitespace removed from the right side.
-|<<functions-sign, sign()>>                     | Numeric           | Returns the signum of a number -- zero if the expression is zero, `-1` for any negative number, and `1` for any positive number.
-|<<functions-sin, sin()>>                       | Trigonometric     | Returns the sine of the expression.
-|<<functions-single,single()>>                  | Predicate         | Returns true if the predicate holds for exactly one of the elements in the list.
+|<<functions-relationships,relationships()/rels()>>    | List              | Returns a list containing all the relationships in a path.
+|<<functions-replace,replace()>>                | String            | Returns a string in which all occurrences of a specified string in the original string have been replaced by another string.
```
alexaverbuch: `"have been replaced with a specified string."`?

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -26,16 +26,16 @@ Most functions in Cypher will return `null` if an input parameter is `null`.
...
 |===
 |Function                         |   Description
-|<<functions-all,all()>>          | Tests whether a predicate holds for all elements in the list.
-|<<functions-any,any()>>          | Tests whether a predicate holds for at least one element in the list.
-|<<functions-exists,exists()>>    | Returns true if a match for the pattern exists in the graph, or the property exists in the node, relationship or map.
-|<<functions-none,none()>>        | Returns true if the predicate holds for no element in the list.
-|<<functions-single,single()>>    | Returns true if the predicate holds for exactly one of the elements in the list.
+|<<functions-all,all()>>          | Tests whether a predicate holds for all elements in a list.
+|<<functions-any,any()>>          | Tests whether a predicate holds for at least one element in a list.
+|<<functions-exists,exists()>>    | Returns true if a match for a pattern exists in the graph, or if a specified property exists in the node, relationship or map.
```
alexaverbuch: As in other feedback

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -47,21 +47,21 @@ These functions return a single value.
...
 | Function                                                            | Description
-| <<functions-coalesce,coalesce()>>                                   | Returns the first non-`null` value in the list of expressions passed to it.
-| <<functions-endnode,endNode()>>                                     | Returns the last node of a relationship.
+| <<functions-coalesce,coalesce()>>                                   | Returns the first non-`null` value in a list of expressions.
+| <<functions-endnode,endNode()>>                                     | Returns the end node of a relationship.
 | <<functions-head,head()>>                                           | Returns the first element in a list.
-| <<functions-id,id()>>                                               | Returns the id of the relationship or node.
+| <<functions-id,id()>>                                               | Returns the id of a relationship or node.
 | <<functions-last,last()>>                                           | Returns the last element in a list.
 | <<functions-length,length()>>                                       | Returns the length of a path.
```
alexaverbuch, line 55: This wasn't part of this PR so feel free to ignore, but doesn't `length()` also work on lists?

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
alexaverbuch: +1

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -74,16 +74,16 @@ These functions take multiple values as arguments, and calculate and return an a
...
 |Function                                       | Description
-|<<functions-avg,avg()>>                        | Returns the average of a numeric column.
-|<<functions-collect,collect()>>                | Returns a list containing all collected values.
-|<<functions-count,count()>>                    | Returns the number of rows.
-|<<functions-max,max()>>                        | Returns the highest value in a numeric column.
-|<<functions-min,min()>>                        | Returns the lowest value in a numeric column.
-|<<functions-percentilecont,percentileCont()>>  | Returns the percentile of a given value over a group using linear interpolation.
+|<<functions-avg,avg()>>                        | Returns the average of a set of numeric values.
+|<<functions-collect,collect()>>                | Returns a list containing the values returned by an expression.
+|<<functions-count,count()>>                    | Returns the number of values or rows.
```
alexaverbuch: shouldn't this be `"... non-null values or rows"`?

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
alexaverbuch: perhaps `"numeric expression"`? though I don't mind number either, because I don't think "number" implies "literal number"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -40,14 +40,15 @@ class AggregatingFunctionsTest extends DocumentingTest {
         |       (a)-[:KNOWS]->(b),
         |       (c)-[:KNOWS]->(d2),
         |       (b)-[:KNOWS]->(d2)""")
-    p("""To calculate aggregated data, Cypher offers aggregation, much like SQL's `GROUP BY`.""")
+    p("""To calculate aggregated data, Cypher offers aggregation, analogous to SQL's `GROUP BY`.""")
     p(
-      """Aggregate functions take multiple input values and calculate an aggregated value from them.
-        |Examples are `avg()` that calculates the average of multiple numeric values, or `min()` that finds the smallest numeric value in a set of values.""")
+      """Aggregating functions take multiple input values and calculate an aggregated value over them.
```
alexaverbuch: `"take multiple input values"` <-- we say `"set of values"` in many other places

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -40,14 +40,15 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        |Examples are `avg()` that calculates the average of multiple numeric values, or `min()` that finds the smallest numeric value in a set of values.
```
alexaverbuch: `min()` does not only apply to numeric values

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -40,14 +40,15 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        |When we say below that an aggregating function operates on a _set of values_, we mean these to be the result of the application of the inner expression (such as `n.age`) to all the records within the same aggregation group.""")
     p(
-      """Aggregation can be done over all the matching subgraphs, or it can be further divided by introducing key values.
+      """Aggregation can be computed over all the matching subgraphs, or it can be further divided by introducing key values.
```
alexaverbuch: is `"key values"` as clear as `"grouping keys"`?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -58,13 +59,13 @@ class AggregatingFunctionsTest extends DocumentingTest {
       """We have two return expressions: `n`, and `count(*)`.
         |The first, `n`, is not an aggregate function, and so it will be the grouping key.
         |The latter, `count(*)` is an aggregate expression.
-        |So the matching subgraphs will be divided into different buckets, depending on the grouping key.
-        |The aggregate function will then run on these buckets, calculating the aggregate values.""")
+        |The matching subgraphs will be divided into different buckets, depending on the grouping key.
+        |The aggregate function will then be run on these buckets, calculating the aggregate values.""")
```
alexaverbuch: `"calculating the aggregate values."` --> `"calculating an aggregate value per bucket"`?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
alexaverbuch: Although I think we consistently and clearly use the terms `list` (to mean a list) & `set of values` (to mean column values) in different ways to each other, I do tend to agree with @craigtaverner that `"set of values"` does not make it clear (at least to me) that we are discussing values of a given columns

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
alexaverbuch: If we change this, it needs to be changed everywhere that `"set of values"` is used.
My other thoughts on that matter are scattered across my other comments in this review :)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/MathematicalNumericFunctionsTest.scala
```diff
@@ -63,18 +64,20 @@ class MathematicalNumericFunctionsTest extends DocumentingTest {
       }
     }
     section("ceil()", "functions-ceil") {
-      p("`ceil()` returns the smallest integer greater than or equal to the argument.")
-      function("ceil(expression)", ("expression", "A numeric expression."))
+      p("`ceil()` returns the smallest floating point number greater than or equal to a given number, and which is equal to an integer.")
```
alexaverbuch: `"and which is equal to an integer."`
is "equal" the correct word here?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -81,42 +81,42 @@ class PredicateFunctionsTest extends DocumentingTest {
...
+      p("`exists()` returns true if a match for a given pattern exists in the graph, or if a specified property exists in the node, relationship or map.")
+      function("exists(pattern-or-property)", "A Boolean.", ("pattern-or-property", "A pattern or a property (in the form 'variable.prop')."))
       query(
         """MATCH (n)
           |WHERE exists(n.name)
           |RETURN n.name AS name, exists((n)-[:MARRIED]->()) AS is_married""".stripMargin, ResultAssertions((r) => {
           r.toList should equal(List(Map("name" -> "Alice", "is_married" -> false), Map("name" -> "Bob", "is_married" -> true), Map("name" -> "Charlie", "is_married" -> false), Map("name" -> "Daniel", "is_married" -> false), Map("name" -> "Eskil", "is_married" -> false)))
         })) {
-        p("This query returns the names of all nodes with a name property along with a boolean `true` / `false` indicating if they are married.")
+        p("The names of all nodes with a `name` property is returned, along with a boolean `true` / `false` indicating if they are married.")
```
alexaverbuch: `"... property is returned, along with a ..."` --> `"... property ARE returned, along with a ..."`?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ScalarFunctionsTest.scala
```diff
@@ -168,19 +173,20 @@ class ScalarFunctionsTest extends DocumentingTest {
...
         resultTable()
-        p("The number of items in the list is returned by the query.")
+        p("The number of elements in the list is returned.")
       }
     }
-    section("Size of pattern expression", "functions-size-of-pattern-expression") {
+    section("size() applied to pattern expression", "functions-size-of-pattern-expression") {
       p(
-        """This is the same `size()` method described before, but instead of passing in a list directly, you provide a pattern expression that can be used in a match query to provide a new set of results.
+        """This is the same `size()` method as described above, but instead of passing in a list directly, pattern expression can be provided that can be used in a match query to provide a new set of results.
```
alexaverbuch: replace `"...  pattern expression can be provided that can be used in ..."`
with `"...  A pattern expression can be provided that can be used in ..."`?

alexaverbuch (changes requested):

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
petraselmer: 1) Values or rows: this is because `count` can operate in two variants: a) (matching) rows, and b) a set of values. This is explained in more detail in the description of `count`, but I had to keep it "short and sweet" here in the Glossary.

2) I think I originally mentioned `non-null` (and also in other places), but decided to be much more brief here (where a one-sentence line was necessary). However, in the description, there is a  new "Considerations" table, which talks about all the facets pertaining to `null` and `count`.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
petraselmer: Well spotted! 👍

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
petraselmer: Will change it to "have been replaced by (another) specified string" - preventing possible reader confusion with the first instance of "specified string"

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
petraselmer: Have added a card: you are absolutely correct 👍

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
petraselmer: Please see comment for this above (all the details pertaining to `null` are in the main description section)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
petraselmer: Good catch!

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
petraselmer: Another good catch 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
petraselmer: It isn't, and I'll change it accordingly 👍  (this entire section needs to be rewritten better  - but that is a later card for a later day)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/MathematicalNumericFunctionsTest.scala
petraselmer: Fixed as discussed

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ScalarFunctionsTest.scala
petraselmer: 👍

petraselmer:
@mariascharin this has been reviewed and discussed with Alex, and, as a result, this PR is now ready to be merged.

PR #122, opened 2017-09-07
Update docs after refactoring

pontusmelke:
If we merge https://github.com/neo4j/neo4j/pull/9986 we need to update docs with this PR


jjaderberg (approved):
@pontusmelke Fire at will.

PR #127, opened 2017-09-26
3.2 procs for schema & reverse(list)

petraselmer:
@mariascharin this is ready to be merged as soon as this PR has been merged and fwd-merged (there are changes in the latter that this PR requires): https://github.com/neo4j/neo4j-documentation/pull/116

For this reason, I have put on the 'DO NOT MERGE' label: this can be removed as soon as PR 116 has been merged.

PR #132, opened 2017-10-03
Added deprecations, additions, removals and comp. info for 3.3

petraselmer:
This must be forward-merged too.
This has a complex dependency chain.

The following PRs need to be merged (and fwd-merged) before this PR can be merged (and a lot of rebasing will ensue):

neo-technology/neo4j-manual-modeling#336

#129 (this deletes the old compat. files and adds the new file)

#127 (depends on the reverse function for lists)

https://github.com/neo4j/neo4j-documentation/pull/130 (this is the 3.2 version of this work, which also requires the above PRs to be in place)

Moreover, although this PR purportedly adds the "deprecations-additions....asciidoc" file, this ought already to be in existence thanks to PR 129. Therefore, what ought to happen here is that the contents of the "deprec..." file ought to be copied **in its entirety** to the version here.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -84,7 +84,7 @@ Replacement syntax for deprecated and removed features are also indicated.
 [options="header"]
 |===
 | Feature          | Type | Change | Details
-| `START`     | Clause | Removed | Replaced by the built-in procedures `db.nodeManualIndexSeek`, `db.nodeManualIndexSearch`, `db.relationshipManualIndexSeek`, and `db.relationshipManualIndexSearch`
+| `START`     | Clause | Removed | As in Cypher 3.2, any queries using the `START` clause will revert back to Cypher 3.1 `planner=rule`. However, there are now built-in procedures for accessing explicit indexes that will enable users to use the current version of Cypher and the cost planner together with these indexes. An example of this is `CALL db.index.explicit.searchNodes('my_index','email:me*')`.
 | `CYPHER runtime=slotted` (Faster interpreted runtime) | Functionality | Added | Neo4j Enterprise Edition only
 | <<functions-max, max()>>, <<functions-min, min()>> | Function  | Extended | Now also supports aggregation over a set containing lists of strings and numbers
 |===
@@ -100,17 +100,18 @@ There are two ways to select which version to use in queries.
 You can configure your database with the configuration parameter `cypher.default_language_version`, and enter which version you'd like to use (see <<cypher-versions>>).
 Every Cypher query will use this version, provided the query hasn't explicitly been configured as described in the next item below.
 
-. Setting a version on a query-by-query basis:
+. Setting a version on a query by query basis:
 The other method is to set the version for a particular query.
 Prepending a query with `CYPHER 2.3` will execute the query with the version of Cypher included in Neo4j 2.3.
 
-Below is an example using the `START` clause to access a legacy index:
+Below is an example using the `has()` function:
 
 [source, cypher]
 ----
 CYPHER 2.3
-START n=node:nodes(name = "A")
-RETURN n
+MATCH (n:Person)
+WHERE has(n.age)
+RETURN n.name, n.age
 ----
 
 
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -85,6 +85,7 @@ Replacement syntax for deprecated and removed features are also indicated.
 |===
 | Feature          | Type | Change | Details
 | `START`     | Clause | Removed | Replaced by the built-in procedures `db.nodeManualIndexSeek`, `db.nodeManualIndexSearch`, `db.relationshipManualIndexSeek`, and `db.relationshipManualIndexSearch`
+| `CYPHER runtime=slotted` (Faster interpreted runtime) | Functionality | Added | Neo4j Enterprise Edition only
 | <<functions-max, max()>>, <<functions-min, min()>> | Function  | Extended | Now also supports aggregation over a set containing lists of strings and numbers
 |===
 
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -11,6 +11,7 @@ New features get added to the language continuously, and occasionally, some feat
  ** <<cypher-deprecations-additions-removals-3.0, Version 3.0>>
  ** <<cypher-deprecations-additions-removals-3.1, Version 3.1>>
  ** <<cypher-deprecations-additions-removals-3.2, Version 3.2>>
+ ** <<cypher-deprecations-additions-removals-3.3, Version 3.3>>
 * <<cypher-compatibility, Compatibility>>
 * <<cypher-versions, Supported language versions>>
 
@@ -78,6 +79,16 @@ Replacement syntax for deprecated and removed features are also indicated.
 |===
 
 
+[[cypher-deprecations-additions-removals-3.3]]
+=== Version 3.3
+[options="header"]
+|===
+| Feature          | Type | Change | Details
+| `START`     | Clause | Removed | Replaced by the built-in procedures `db.nodeManualIndexSeek`, `db.nodeManualIndexSearch`, `db.relationshipManualIndexSeek`, and `db.relationshipManualIndexSearch`
+| <<functions-max, max()>>, <<functions-min, min()>> | Function  | Extended | Now also supports aggregation over a set containing lists of strings and numbers
+|===
+
+
 [[cypher-compatibility]]
 == Compatibility
 
@@ -105,8 +116,9 @@ RETURN n
 [[cypher-versions]]
 == Supported language versions
 
-Neo4j 3.2 supports the following versions of the Cypher language:
+Neo4j 3.3 supports the following versions of the Cypher language:
 
+* Neo4j Cypher 3.3
 * Neo4j Cypher 3.2
 * Neo4j Cypher 3.1
 * Neo4j Cypher 2.3


cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+* <<cypher-deprecations-additions-removals, Removals, deprecations, additions and extensions>>
+ ** <<cypher-deprecations-additions-removals-3.0, Version 3.0>>
+ ** <<cypher-deprecations-additions-removals-3.1, Version 3.1>>
+ ** <<cypher-deprecations-additions-removals-3.2, Version 3.2>>
+ ** <<cypher-deprecations-additions-removals-3.3, Version 3.3>>
+* <<cypher-compatibility, Compatibility>>
+ ** <<cypher-start-compatibility, Accessing entities by id via START>>
+* <<cypher-versions, Supported language versions>>
+
+[abstract]
```
systay: I noted that other docs in this folder have the abstract before the index and not after like this. Not important to me, simply pointing it out.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+| <<user-defined-functions, User-defined functions>> | Functionality | Added |
+| <<query-call, CALL\...YIELD\...WHERE>>   | Clause  | Extended  | Records returned by `YIELD` may be filtered further using `WHERE`
+|===
+
+[[cypher-deprecations-additions-removals-3.2]]
+=== Version 3.2
+[options="header"]
+|===
+| Feature          | Type | Change | Details
+| `CYPHER planner=rule` (Rule planner)    | Functionality | Removed | All queries now use the cost planner. Any query prepended thus will fall back to using Cypher 3.1.
```
systay: Not sure I would write "removed" here. If you are using Neo4j 3.2 and specify "planner=rule", that query will still work and return the expected results. I would say deprecated here. Same goes for the next element, `CREATE UNIQUE`.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+There are two ways to select which version to use in queries.
+
+. Setting a version for all queries
+You can configure your database with the configuration parameter `cypher.default_language_version`, and enter which version you'd like to use (see <<cypher-versions>>).
+Every Cypher query will use this version, provided the query hasn't explicitly been configured as described in the next item below.
+
+. Setting a version on a query by query basis
+The other way is on a query by query basis.
+By simply putting `CYPHER 2.3` at the beginning, that particular query will be executed with the version of Cypher included in Neo4j 2.3.
+
```
systay: Maybe point out that if both these settings are set, the pre-parser option will win.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+----
+CYPHER 1.9
+START n=node(42)
+RETURN n
+----
+
+[NOTE]
+The use of the `START` clause to find nodes by ID was deprecated from Cypher 2.0 onwards and is now entirely disabled in Cypher 2.2 and up.
+You should instead make use of the `MATCH` clause for starting points.
+See <<query-match>> for more information on the correct syntax for this.
```
systay: The link for <<query-match>> doesn't work, but that is expected, right?

systay (approved):
Looks great! Minor tweaks suggested

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: Correct (only works when it's all built, and the Docs team check and verify that this is so)

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
craigtaverner: Also the rendered version of the abstract looks block-quoted, which seems wrong to me.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
craigtaverner: This is an odd case. For Neo4j 3.2 we have removed the rule planner, but since we support Cypher 3.1, and that bundles its own rule planner, we can still get it that way. So the question is, does the fact that we coded a convenience of having 'Cypher planner=rule' automatically mean 'Cypher 3.1 planner=rule' mean that we should not use the word 'removed' and instead use the word 'deprecated'?

craigtaverner: We could argue that if the title of this table `Version 3.2` refers to Cypher version, then these features are indeed _removed_ and not merely deprecated. If we are refering to Neo4j the product, then they are deprecated because we automatically switch to and older Cypher.

My vote is for keeping the word '_removed_'.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+| <<functions-reverse-list,reverse()>> | Function  | Extended | Now also allows a list as input
+| <<functions-max, max()>>, <<functions-min, min()>> | Function  | Extended | Now also supports aggregation over a set containing both strings and numbers
+|===
+
+[[cypher-deprecations-additions-removals-3.3]]
+=== Version 3.3
+[options="header"]
+|===
+| Feature          | Type | Change | Details
+| `START`     | Clause | Removed | Replaced by the built-in procedures `db.nodeManualIndexSeek`, `db.nodeManualIndexSearch`, `db.relationshipManualIndexSeek`, and `db.relationshipManualIndexSearch`
```
craigtaverner: This is subtle. We already claimed it was removed in the previous version. We have not removed it more, and the fallback to Cypher 3.1 planner=rule still works, but we have provided new alternatives that allow the use of the Cost planner together with explicit indexes. Also, the names of the procedures are incorrect. Since there are many procedures, let's link to the documentation on the full list from here. If we want to use examples, let's use the correct names.

craigtaverner: What about text like:

> As in Cypher 3.2, any queries using the `START` clause will revert back to `Cypher 3.1 planner=rule`. However, there are now built-in _procedures for accessing explicit indexes_ that will enable users to use the current version of Cypher and the COST planner together with these indexes. For example `CALL db.index.explicit.searchNodes('my_index','email:me*')`.

Make sure _procedures for accessing explicit indexes_ is a link to the correct section.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+
+Older versions of the language can still be accessed if required.
+There are two ways to select which version to use in queries.
+
+. Setting a version for all queries
+You can configure your database with the configuration parameter `cypher.default_language_version`, and enter which version you'd like to use (see <<cypher-versions>>).
+Every Cypher query will use this version, provided the query hasn't explicitly been configured as described in the next item below.
+
+. Setting a version on a query by query basis
+The other way is on a query by query basis.
```
craigtaverner: You have two sentences that duplicate each other and are rendered as one sentence here.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
craigtaverner: That is mentioned in the previous paragraph. The text states:
> Every Cypher query will use this version, provided the query hasn't explicitly been configured as described in the next item below

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+[source, cypher]
+----
+CYPHER 2.3
+START n=node:nodes(name = "A")
+RETURN n
+----
+
+
+[[cypher-start-compatibility]]
+=== Accessing entities by id via START
```
craigtaverner: This entire section seems like content from a Cypher 2.2 or Cypher 2.3 manual. Everything above here is very much about Neo4j 3.0 and later. Do we still need this information about accessing nodes by id using START at all?

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+
+. Setting a version for all queries
+You can configure your database with the configuration parameter `cypher.default_language_version`, and enter which version you'd like to use (see <<cypher-versions>>).
+Every Cypher query will use this version, provided the query hasn't explicitly been configured as described in the next item below.
+
+. Setting a version on a query by query basis
+The other way is on a query by query basis.
+By simply putting `CYPHER 2.3` at the beginning, that particular query will be executed with the version of Cypher included in Neo4j 2.3.
+
+Below is an example using the `START` clause to access a legacy index:
```
craigtaverner: The name 'legacy' should be replaced by 'explicit'. And the example seems appropriate to users of older versions. For 3.3 we would prefer to use the explicit index procedures for accessing those indexes. Perhaps an example of something else that is removed, like `has()`?

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+See <<query-match>> for more information on the correct syntax for this.
+The `START` clause should only be used when accessing explicit indexes.
+
+[[cypher-versions]]
+== Supported language versions
+
+Neo4j 3.3 supports the following versions of the Cypher language:
+
+* Neo4j Cypher 3.3
+* Neo4j Cypher 3.2
```
craigtaverner: We support 3.1 also for the specific reason we need to have the RULE planner for START and CREATE UNIQUE.

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: Yup, I was flip-flopping on this one a lot. I settled on `Removed` for the reason Craig mentioned. So, will keep as is, and hopefully the wording (all about how the query falls back yada yada) is sufficient to make this distinction clear to readers.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: Added the text. The link will be sorted out as per the final Deprecations checking card.

petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: Ah! This was old text (copied from the intro section). Well-spotted - will also make other changes to improve the text.

petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: The changes to rename "legacy" -> "explicit" is still to be undertaken. I will leave the text as-is for 3.1 and 3.2 (I've been applying these changes, where appropriate, to 3.1 and 3.2 as well). However, for 3.3 I will replace this with a `has()` example - you're quite right here.

petraselmer: 👍

petraselmer: Actually, will do for 3.2 as well

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: You're right - esp. in light of the new Deprec/Adds/Rems table, which explains the gist of this anyway. Removing in 3.1, 3.2 and 3.3

petraselmer: 👍

petraselmer:
@craigtaverner I have addressed your feedback

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
petraselmer: 👍

PR #134, opened 2017-10-04
3.1 - Preparser options

petraselmer:
@craigtaverner this is ready for you to look at. Versions 3.2 and 3.3 will mostly be additive, but with extras, so each will will need to be reviewed in turn.

@mariascharin / @jjaderberg / @davidoliverSP2  Docs Team as a whole: a new term needs to be invented/selected/decided upon for "Preparser options". You'll notice I have peppered "TODOs" liberally in the PR. Leaving this to you guys to decide. Thanks!

Just transcribing text from @craigtaverner in case it gets hidden when other textual changes are made: 
Note to team-docs: I think the reason they are called pre-parser options is that they are processed first, and based on them we decide what should do the 'real' parsing, planning and query execution. The fact that this must happen first should be obvious to the user, and yet we could also see this as being an implementation detail, and so an alternative heading might be preferred. However the older heading 'how are queries executed' seems even worse, because it sounds like it is about to describe the details of planning and runtime, which is certainly not what this section is about. What about something like 'Query execution engine selection'?


cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -35,6 +35,33 @@ To read more about the execution plan operators mentioned in this chapter, see <
 [[how-are-queries-executed]]
 == How are queries executed?
 
+_***XXTODO by Docs Team - change this heading to something like 'Preparser options' (this is what they are called internally), or something more reader-friendly (if a term other than 'preparser' is decided upon, please also change accordingly in the text). This section no longer deals with 'how queries are executed' per se. More of that content will be added by me later on. END TODO***_
```
craigtaverner: Note to team-docs: I think the reason they are called pre-parser options is that they are processed first, and based on them we decide what should do the 'real' parsing, planning and query execution. The fact that this must happen first should be obvious to the user, and yet we could also see this as being an implementation detail, and so an alternative heading might be preferred. However the older heading 'how are queries executed' seems even worse, because it sounds like it is about to describe the details of planning and runtime, which is certainly not what this section is about. What about something like 'Query execution engine selection'?

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -35,6 +35,33 @@ To read more about the execution plan operators mentioned in this chapter, see <
...
+
+[abstract]
+--
+This section describes the preparser options available in Cypher.
+--
+
+Queries are able to be fine-tuned through the use of _preparser options_.
```
craigtaverner: 'Query execution can be fine-tuned ...'

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -35,6 +35,33 @@ To read more about the execution plan operators mentioned in this chapter, see <
...
+If there is a requirement to use one or more of these options, the query must be prepended with `CYPHER`, followed by the preparser option(s), as exemplified thus: `CYPHER preparser-option1 [further preparser options] query`.
```
craigtaverner: Perhaps simplified to 'In order to use these options, the query must be prepended with ...'

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -47,8 +74,14 @@ Cost::
...
-By default, Neo4j {neo4j-version} will use the cost planner for all queries.
-You can force it to use a specific planner by using the `cypher.planner` configuration setting (see <<operations-manual#config_cypher.planner, Operations Manual -> Configuration Settings>>), or by prepending your query with `CYPHER planner=cost` or `CYPHER planner=rule`.
+[options="header"]
+|===
+| Preparser option | Description | Default?
+| `planner=rule` | This will force the query to use the rule planner.  |
+| `planner=cost` | This will force the query to use the cost planner. As this is the default planner, it is not necessary to use this option explicitly. | X
+|===
+
+It is also possible to force queries to use a specific planner by using the `cypher.planner` configuration setting (see <<operations-manual#config_cypher.planner, Operations Manual -> Configuration Settings>>).
```
craigtaverner: Is this really a 'force' option? According to the inline docs in GraphDatabaseSettings, this just sets the default planner. We should re-word this accordingly.

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
petraselmer: Transcribed above so that Docs Team can see it all in one... (I like your suggestion, btw)

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
petraselmer: 👍 (slight amendment: wanted to retain and emphasise "one or more")

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

petraselmer:
@mariascharin More comment by Craig on this version, so putting it back under merge embargo

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -47,8 +74,14 @@ Cost::
 This planner uses the statistics service in Neo4j to assign cost to alternative plans and picks the cheapest one.
 While this should lead to superior execution plans in most cases, it is still under development.
```
craigtaverner: We could consider the COST planner complete for reads in 2.3, and complete for writes in 3.0. I think this document simply has not been updated to reflect that fact. We did not materially change the COST planner after 3.0 other than minor fixes or performance enhancements, so the removal of RULE in 3.2 was not because COST in 3.2 was now complete, but because COST had already been complete for a couple of releases.

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -47,8 +74,14 @@ Cost::
...
-By default, Neo4j {neo4j-version} will use the cost planner for all queries.
-You can force it to use a specific planner by using the `cypher.planner` configuration setting (see <<operations-manual#config_cypher.planner, Operations Manual -> Configuration Settings>>), or by prepending your query with `CYPHER planner=cost` or `CYPHER planner=rule`.
+[options="header"]
+|===
+| Preparser option | Description | Default?
+| `planner=rule` | This will force the query to use the rule planner.  |
+| `planner=cost` | This will force the query to use the cost planner. As this is the default planner, it is not necessary to use this option explicitly. | X
+|===
+
+It is also possible to set the default planner by using the `cypher.planner` configuration setting (see <<operations-manual#config_cypher.planner, Operations Manual -> Configuration Settings>>).
```
craigtaverner: Perhaps the word 'set' should be changed to 'change'? Reads better to me.

petraselmer:
@mariascharin Now ready to merge

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

mariascharin:
Note that PR https://github.com/neo-technology/neo4j-manual-modeling/pull/347 must be merged together with this one!

PR #133, opened 2017-10-03
3.2 Fixes: removal of RULE planner, START and CREATE UNIQUE workarounds; Preparser options

petraselmer:
@craigtaverner the preparser options information is ready for your review. There are quite a few differences from 3.1.

@mariascharin This can only be merged once reviewed by Craig. However, this PR must be  merged first: https://github.com/neo4j/neo4j-documentation/pull/134

Moreover, please note that there is a lot of old compatibility content which was removed in another PR. Thus, when merging this, take as truth the main codebase when it comes to any compatibility-related errors.

Moreover - Docs Team as a whole - a new term needs to be invented/selected/decided upon for "Preparser options", just as for the 3.1 version. You'll notice I have peppered "TODOs" liberally in the PR. Additionally, sub-pages need to be created under the new "Deprecations" section (not merged at the time of raising this PR, hence not done in conjunction with this PR) for `START` and `CREATE UNIQUE`

PR #135, opened 2017-10-05
3.3 - Added preparsing options details

petraselmer:
@craigtaverner the preparser options information is ready for your review. There are quite a few differences from 3.1 and 3.2. In particular, there is still the issue of supported versions vs compatibility vs 3.1 fallbacks (due to removal of the RULE planner in 3.2)

@mariascharin This can only be merged once reviewed by Craig. However, this PR needs to be merged before this one: https://github.com/neo4j/neo4j-documentation/pull/133

Moreover - Docs Team as a whole - a new term needs to be invented/selected/decided upon for "Preparser options", just as for the 3.2 and 3.1 version. You'll notice I have peppered "TODOs" liberally in the PR.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
@@ -103,6 +103,7 @@ Slotted::
 This is very similar to the interpreted runtime, except that there are additional optimizations regarding the way in which the records are streamed through the iterators.
 This results in improvements to both the performance and memory usage of the query.
 In effect, this can be thought of as a the 'faster interpreted' runtime.
+This is still under development and does not support all possible operators or queries.
 
 [NOTE]
 The slotted runtime is only available in Neo4j Enterprise Edition.
@@ -114,12 +115,15 @@ While this should lead to superior performance in most cases (over both the inte
 [NOTE]
 The compiled runtime is only available in Neo4j Enterprise Edition.
 
+[NOTE]
+In Enterprise Edition, the query execution engine's fallback mechanism is to try the compiled runtime first, and if it does not support the query, to then try the slotted runtime, and if that also does not support the query, to finally fall back to the interpreted runtime, which supports all queries.
+
 [options="header"]
 |===
 | Option | Description | Default?
 | `runtime=interpreted` | This will force the query execution engine to use the interpreted runtime. | This is the default option for some queries in Enterprise Edition, and the only option for all queries in Community Edition (so it is not necessary to use this option in Community Edition).
-| `runtime=slotted` | This will force the query to use the slotted runtime. | This is the default option for some queries in Enterprise Edition only. Setting this option for a query that does not support the slotted runtime will cause the query to fall back to using the interpreted runtime.
-| `runtime=compiled` | This will cause the query execution engine to use the compiled runtime if it supports the query, and fall back to using the interpreted runtime if it does not. | This is the default option for some queries in Enterprise Edition only.
+| `runtime=slotted` | This will cause the query execution engine to use the slotted runtime if it supports the query, and fall back to using the interpreted runtime if it does not. | This is the default option for some queries in Enterprise Edition only.
+| `runtime=compiled` | This will cause the query execution engine to use the compiled runtime if it supports the query. If the compiled runtime does not support the query, the engine will fall back to the slotted runtime, and, failing that, to the interpreted runtime. | This is the default option for some queries in Enterprise Edition only.
 |===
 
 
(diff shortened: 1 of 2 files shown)


cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -35,20 +35,53 @@ To read more about the execution plan operators mentioned in this chapter, see <
...
+
+Occasionally, there is a requirement to use a previous version of the Cypher compiler when running a query.
+Here we detail the available versions:
+
+[options="header"]
+|===
+| Preparser option | Description | Default?
+| `2.3` | This will force the query to use Neo4j Cypher 2.3.  |
+| `3.2` | This will force the query to use Neo4j Cypher 3.2. |
+| `3.1` | This will force the query to use Neo4j Cypher 3.1. |
```
craigtaverner: Odd order, perhaps put 3.1 above 3.2

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -35,20 +35,53 @@ To read more about the execution plan operators mentioned in this chapter, see <
...
 
-Rule::
-This planner has rules that are used to produce execution plans.
-The planner considers available indexes, but does not use statistical information to guide the query compilation.
+Neo4j uses a _cost_-based execution planning strategy (known as the 'cost' planner): the statistics service in Neo4j is used to assign a cost to alternative plans and picks the cheapest one.
 
-Cost::
-This planner uses the statistics service in Neo4j to assign cost to alternative plans and picks the cheapest one.
-While this should lead to superior execution plans in most cases, it is still under development.
+All versions of Neo4j prior to Neo4j 3.2 also included a rule-based planner, which used rules to produce execution plans.
```
craigtaverner: I suggested simpler text in the 3.1 review.

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -57,6 +90,40 @@ When Cypher is building execution plans, it looks at the schema to see if it can
...
+This is very similar to the interpreted runtime, except that there are additional optimizations regarding the way in which the records are streamed through the iterators.
+This results in improvements to both the performance and memory usage of the query.
+In effect, this can be thought of as a the 'faster interpreted' runtime.
+
+[NOTE]
+The slotted runtime is only available in Neo4j Enterprise Edition.
+
+Compiled::
+Algorithms are employed to group intelligently the operators in the execution plan in order to generate new combinations and orders of execution which are optimised for performance and memory usage.
+While this should lead to superior performance in most cases (over both the interpreted and slotted runtimes), it is still under development.
```
craigtaverner: I made a comment in the 3.1 review about this not supporting all operators or queries. But both that and what is already written here is also true of slotted. We could describe that both compiled and slotted provide performance advantages, and both do not cover all operators and queries, but while compiled has the highest performance advantage, it also has the lowest coverage. So the fallback mechanism in enterprise is to try compiled first, and if it does not support the query, then try slotted, and if that also does not support the query finally fall back to interpreted which supports all queries. In community, only interpreted is considered.

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -57,6 +90,40 @@ When Cypher is building execution plans, it looks at the schema to see if it can
...
+
+[NOTE]
+The compiled runtime is only available in Neo4j Enterprise Edition.
+
+[options="header"]
+|===
+| Preparser option | Description | Default?
+| `runtime=interpreted` | This will force the query to use the interpreted runtime.  | This is the default option for some queries in Enterprise Edition, and the only option for all queries in Community Edition (so it is not necessary to use this option in Community Edition).
+| `runtime=slotted` | This will force the query to use the slotted runtime. | This is the default option for some queries in Enterprise Edition only. Setting this option for a query that does not support the slotted runtime will cause the query to fall back to using the interpreted runtime.
+| `runtime=compiled` | This will force the query to use the compiled runtime. | This is the default option for some queries in Enterprise Edition only. Setting this option for a query that does not support the compiled runtime will cause the query to fall back to using the slotted or interpreted runtime.
```
craigtaverner: As in my 3.2 comments, I think the word 'force' is too strong for the fallback process. I would prefer that we have a separate paragraph describing the fallback process, and then in this table only describe how these options simply set at which level in that process to start.

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
petraselmer: Also responded in that review (3.2)

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
petraselmer: 👍 👍

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

PR #145, opened 2017-10-12
Explicit indexes; fix typo

petraselmer:


jjaderberg:
@petraselmer should this target 3.3 rather than 3.4?

petraselmer:
@jjaderberg it should indeed (target 3.3) - good catch! I'll wait for @craigtaverner to provide his feedback, address it and then re-open...

petraselmer:
Replaced by https://github.com/neo4j/neo4j-documentation/pull/146

PR #160, opened 2017-11-22
Fixed dependencies due to kernel API changes

craigtaverner:
When trying to compile locally, I found these changes were required based on recent kernel API changes.


craigtaverner:
Aparently an issue with my computer only

PR #162, opened 2017-11-22
Explicit indexing procedures docs

jjaderberg:


cypher/cypher-docs/src/docs/dev/ql/schema-index/explicit-index-procedures.adoc
```diff
@@ -1,34 +1,219 @@
...
-| |
+|===
+
+
+[[db.index.explicit.searchRelationships]]
+.db.index.explicit.searchRelationships
+[options="header"]
+|===
+| Signature | Description
+|
```
mariascharin, line 173: This detail table is empty. Copy/paste mistake?

mariascharin (commented):
Wow, this is a huge improvement!
I only have a detail observation that the width of the columns in the detail tables isn't consistent, which gives a "wavy" feeling when you view the page.

jjaderberg:
Ta 👍 
The varying width in the details tables is a bit jarring, it's the same but less extreme in the settings reference. I don't have a good way to relieve that at the moment, but we can keep it in mind going forward.

PR #165, opened 2017-12-07
More examples for UNWIND

petraselmer:


petraselmer:
@craigtaverner please review. If all ok, please remove the DO NOT MERGE label

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UnwindTest.scala
```diff
@@ -28,26 +28,76 @@ class UnwindTest extends DocumentingTestBase {
...
+        "The second `UNWIND` then operates on each of these rows in turn, resulting in five rows for `y`.",
+      assertions = (p) => assertEquals(List(1, 2, 3, 4, 5), p.columnAs[Int]("y").toList)
+    )
+  }
+
+  @Test def empty_list() {
+    testQuery(
+      title = "Using `UNWIND` with an empty list",
+      text = "Using an empty list with `UNWIND` will produce no rows, irrespective of whether or not any rows existed beforehand, or whether or not other values are being projected. " +
+        "Essentially, `UNWIND []` reduces the number of rows to zero, and thus causes the query to cease its execution, returning no results.",
```
craigtaverner: It is super valuable to explain the behaviour of `UNWIND []`, but it leads also to the question of why someone would ever do this. or they might as well not write the query at all. Of course, they would not write `UNWIND []`, but rather `UNWIND x` where x is some previous variable, or parameter that may or may not by an empty list at runtime. So it might be worth adding a little more to the description to explain how this feature can provide value.

Perhaps just adding to the end of the description: This has value in cases like `UNWIND v` where `v` is a variable from an earlier clause that may or may not be an empty list, and when it is an empty list this will behave just as a `MATCH` that has no results.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UnwindTest.scala
```diff
@@ -28,26 +28,76 @@ class UnwindTest extends DocumentingTestBase {
...
+        """UNWIND [] AS empty
+           RETURN empty, 'literal_that_is_not_returned'""".stripMargin,
+      assertions = (p) => assertEquals(List(), p.columnAs[Int]("empty").toList)
+    )
+  }
+
+  @Test def using_non_lists() {
+    testQuery(
+      title = "Using `UNWIND` with an expression that is not a list",
+      text = "Attempting to use `UNWIND` on an expression that does not return a list will cause an error. " +
```
craigtaverner: Two concerns with this test:

* The test claims to test `using_non_list`, but in fact tests `unwind on null`, which is a related concept. I think perhaps two tests, one for null and one for a non-null, non-list.
* Is it really true that a non-list gives an error? I thought `UNWIND 5` would be the same as `UNWIND [5]`, and there is an example above in `nested_lists` that demonstrates that behaviour.

A third comment for the CLG is why null behaves differently to other values, when `UNWIND [null]` should return a single row containing null, and `UNWIND 5' should return a single row containing 5, why should `UNWIND null` not return a single row containing null?

craigtaverner (dismissed):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UnwindTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UnwindTest.scala
petraselmer: I have amended the description to include a mention of how something along the lines of `UNWIND 5` is an error. This is actually a bug, and the CLG have it on the wall (raised by Edinburgh as a result of the formal semantics work). When the bug is resolved - and how this will look remains to be decided - the docs will change accordingly

petraselmer:
@mariascharin this is  now ready for merging

InverseFalcon:
Whoops, I think one of the examples won't execute:

```
WITH [[1, 2], [3, 4], 5] AS nested
          UNWIND nested AS x
          UNWIND x AS y
          RETURN y
```

We get an error because we're attempting to unwind 5

InverseFalcon:
Also for using UNWIND on an empty list, it may be worth adding a comment that you can use CASE to replace an empty list with a list of `null` for an UNWIND if you want to avoid wiping out the row:

```
WITH [] as list
UNWIND CASE WHEN list = [] THEN [null] ELSE list END  AS empty
```
This will result in a single row where `empty` is `null`.

mariascharin:
@petraselmer Did you see the comments on this?

petraselmer:
@mariascharin yes. I will raise a new PR for 3.2 ->

PR #167, opened 2018-01-04
Execution plan operators: total revamp

petraselmer:
Once this has been merged, merge https://github.com/neo-technology/neo4j-manual-modeling/pull/416


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -118,7 +118,7 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Distinct",
       text =
-        """Removes duplicate rows from the incoming stream of rows.""".stripMargin,
+        """The `Distinct` operator removes duplicate rows from the incoming stream of rows.""".stripMargin,
```
systay: This operator needs to build up state to make sure only distinct elements are returned. This leads to added memory pressure in the system.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -164,11 +164,11 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Eager",
       text =
-        """For isolation purposes this operator makes sure that operations that affect subsequent operations are executed fully for the whole dataset before continuing execution.
-           | Otherwise it could trigger endless loops, matching data again, that was just created.
+        """For isolation purposes, the `Eager` operator ensures that operations affecting subsequent operations are executed fully for the whole dataset before continuing execution.
+           | Otherwise, endless loops could be triggered in which data that was just created is matched.
```
systay: > Otherwise, endless loops could be triggered in which data that was just created is matched.

should be replaced with something like:

> Information from the stores is fetched in a lazy manner, i.e. the pattern matching might not be fully exhausted before we start applying updates. To keep the semantics sane, the query planner will insert Eager operators into the query plan to prevent updates to influence pattern matching. This example query for example, where the DELETE will influence the MATCH.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -209,23 +209,51 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+               |If the index is not unique, the operator is instead called `NodeIndexSeek`.""".stripMargin,
+      queryText = """MATCH (t:Team {name: 'Malmo'}) RETURN t""",
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("NodeUniqueIndexSeek"))
+    )
+  }
+
+  @Test def argument() {
+    profileQuery(
+      title = "Argument",
+      text = """The `Argument` operator indicates the variable to be used as an argument to the right-hand side of an <<query-plan-apply, Apply>> operator.""".stripMargin,
```
systay, line 521: maybe make it clear that it could also be one of the many siblings and cousins of Apply?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -128,7 +128,7 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Eager Aggregation",
       text =
-        """Eagerly loads underlying results and stores it in a hash-map, using the grouping keys as the keys for the map.""".stripMargin,
+        """The `EagerAggregation` operator eagerly loads underlying results and stores it in a hash map, using the grouping keys as the keys for the map.""".stripMargin,
```
systay: This is a very mechanical explanation for what is happening. What about:

The `EagerAggregation` operator evaluates the grouping expressions and uses the result to group rows into different groupings. For each of these groupings, this operator will then evaluate all aggregation functions and return the result. 
To do this, the operator, as the name implies, needs to eagerly pull in all data from its source and hold it in the query state, which leads to added memory pressure in the system.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -209,23 +209,51 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+      text = """The `Argument` operator indicates the variable to be used as an argument to the right-hand side of an <<query-plan-apply, Apply>> operator.""".stripMargin,
+      queryText = """MATCH (s:Person {name: 'me'}) MERGE (s)-[:FRIENDS_WITH]->(s)""",
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("Argument"))
+    )
+  }
+
+  @Test def loadCSV() {
+    profileQuery(
+      title = "Load CSV",
+      text = """The `LoadCSV` operator is used when executing a query containing the <<query-load-csv, LOAD CSV>> clause.""".stripMargin,
```
systay: The LoadCSV operator loads data into the query from a CSV source. It is used whenever the LOAD CSV clause is used in a query.

// I just thought it to be good practise to talk about what the operator actually does and not only which kinds of queries would use it. Just a detail, but still :)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -209,23 +209,51 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+      queryText = """LOAD CSV FROM 'https://neo4j.com/docs/cypher-refcard/3.3/csv/artists.csv' AS line RETURN line""",
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("LoadCSV"))
+    )
+  }
+
   @Test def nodeIndexRangeSeek() {
     executePreparationQueries {
       val a = (0 to 100).map { i => "CREATE (:Location)" }.toList
```
systay: I don't think this extra creation of nodes with the Location label but without the name property is needed anymore. This probably applies to all indexing operator examples.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -235,22 +263,43 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+
   @Test def nodeIndexScan() {
     executePreparationQueries((0 to 250).map { i =>
       "CREATE (:Location)"
     }.toList)
-    profileQuery(title = "Node index scan",
+    profileQuery(title = "Node Index Scan",
                  text = """
-                          |An index scan goes through all values stored in an index, and can be used to find all nodes with a particular label having a specified property (e.g. `exists(n.prop)`).""".stripMargin,
+                          |The `NodeIndexScan` operator goes through all values stored in an index, and searches for all nodes with a particular label having a specified property (e.g. `exists(n.prop)`).""".stripMargin,
```
systay: "searching for" is not really correct - the index only contains nodes that match this description, so it's more accurate to say that it "returns all nodes with a particular..."

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -315,7 +379,7 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Optional Expand All",
       text =
-        """Optional expand traverses relationships from a given node, and makes sure that predicates are evaluated before producing rows.
+        """The `OptionalExpand(All)` operator traverses relationships from a given node, and ensures that predicates are evaluated before producing rows.
```
systay: This could read something like:

OptionalExpand(All) does the same thing as Expand(All) does, except when no relationships match the direction, type and property predicates. In that case, OptionalExpand(all) will return a single row with the relationship and end node are set to `null`

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -330,7 +394,7 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Sort",
       text =
-        """Sorts rows by a provided key.""".stripMargin,
+        """The `Sort` operator sorts rows by a provided key.""".stripMargin,
```
systay: To be able to do this, it needs to eagerly pull in all data from its source operator and keep it query state before being able to actually sort the data. This leads to added memory pressure.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -370,17 +464,49 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Expand Into",
       text =
-        """When both the start and end node have already been found, expand-into is used to find all connecting relationships between the two nodes.""".stripMargin,
+        """When both the start and end node have already been found, the `Expand(Into)` operator is used to find all relationships connecting the two nodes.""".stripMargin,
```
systay: Since both sides of the relationship are already in scope, the side with the smallest degree will be used. This can make a big difference when dense nodes can show up as end points.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -370,17 +464,49 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
     )
   }
 
+
+  @Test def optionalExpandInto() {
+    profileQuery(
+      title = "Optional Expand Into",
+      text =
+        """When both the start and end node have already been found, the `OptionalExpand(Into)` operator is used to find all relationships connecting the two nodes.
+          |If no matching relationships are found, a single row with `null` for the relationship and end node variable is produced.xxxxx""".stripMargin,
```
systay: Maybe also: Since both sides of the relationship are already in scope, the side with the smallest degree will be used. This can make a big difference when dense nodes can show up as endpoints.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -1048,7 +1048,9 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Node Outer Hash Join",
       text =
-        """The `NodeOuterHashJoin` operator ....""".stripMargin,
+        """Using a hash table, the `NodeOuterHashJoin` operator joins the input coming from the left with the input coming from the right.
+          |If the input from the left does not have any matches coming from the right, a `null` is produced for the variable on the right.
+          |`NodeOuterHashJoin` only gets planned for larger cardinalities; for smaller cardinalities, `Expand` is used instead.""".stripMargin,
```
systay: This is not strictly true. It's just a cost model calculation - pretty sure we could trigger it with quite low cardinalities.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -438,12 +438,22 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
       queryText = """CREATE (:Person)""",
       assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("EmptyResult"))
     )
   }
 
+  @Test def produceResult() {
+    profileQuery(
+      title = "Produce Result",
+      text =
+        """The `ProduceResult` operator ....""".stripMargin,
```
systay: The `ProduceResult` operator prepares the result in a way that is consumable by the user - transforming internal values to user values. It's present in every single query that returns data to the user, and not something that usually shows up during performance optimisation.

Not sure into how much detail we should go in here...

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ShortestPathPlanningTest.scala
```diff
@@ -109,6 +109,10 @@ class ShortestPathPlanningTest extends DocumentingTest {
             |stream.
             |On top of this, the planner will issue an `AntiConditionalApply`, which will run the exhaustive search
             |if the path variable is pointing to `null` instead of a path.""")
+        p(
+          """An `ErrorPlan` operator will appear in the execution plan in cases where (i)
+            |`cypher.forbid_exhaustive_shortestpath` is set to `true`, and (ii) the fast algorithm is not able to fulfill the query.""".stripMargin
```
craigtaverner: Looks good, although _not able to fulfill the query_ might be misleading. The real thing is that shortestPath will give no results, and we cannot tell if that really means there are no shortestPaths, or whether it just failed to find any conforming to the predicates. Perhaps change the text to 'the fast algorithm was not able to find a shortestPath'. The difference is subtle, I think.

craigtaverner (dismissed):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ShortestPathPlanningTest.scala
petraselmer: I have pushed this change 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -211,7 +211,8 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
     profileQuery(
       title = "Distinct",
       text =
-        """The `Distinct` operator removes duplicate rows from the incoming stream of rows.""".stripMargin,
+        """The `Distinct` operator removes duplicate rows from the incoming stream of rows.
+          |To ensure only distinct elements are returned, `Distinct` needs to eagerly pull in all data from its source and build up state, which will lead to increased memory pressure in the system.""".stripMargin,
```
systay: I've double checked the code again and realised that what I suggested this to say is not true. DISTINCT does not eagerly load anything. It does build up the same state, but lazily. So, it's still not great from a memory consumption POV, but less awful than eager loading at least.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -441,16 +836,287 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+      queryText = """UNWIND range(1, 5) as value return value""",
       assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("Unwind"))
     )
   }
 
+  @Test def apply() {
+    profileQuery(
+      title = "Apply",
+      text =
+        """`Apply` works by performing a nested loop.
```
systay: All the different `Apply` operators share the same basic functionality - they perform a nested loop, taking a single row from the LHS, and through the `Argument` operator on the RHS, executes the RHS subtree. What happens with these results is what the variation between the different `Apply` versions. 

The unadorned `Apply` operator takes the row produced from the RHS, which by now contains data from both sides, and yields it.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -441,16 +836,287 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+      title = "Anti Conditional Apply",
+      text =
+        """The `AntiConditionalApply` operator checks whether a variable is `null`, and if so, the right child operator will be executed.""".stripMargin,
+      queryText =
+        """MERGE (p:Person {name: 'Andres'})
+          |ON CREATE SET p.exists = true""".stripMargin,
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("AntiConditionalApply"))
+    )
+  }
+
```
systay, line 1005: `RollUpApply`

The RollUpApply is used to execute expressions that take a pattern and returns a list with content from the matched pattern, e.g. pattern expressions or pattern comprehension.

queryTest = "MATCH (p:Person) RETURN p.name, [ (p)-[:WORKS_IN]->(location) | location.name ] as cities"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -441,16 +836,287 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+          |RETURN other.name""".stripMargin,
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("Triadic"))
+    )
+  }
+
+  @Test def foreach() {
+    profileQuery(
+      title = "Foreach",
+      text =
+        """The `Foreach` operator xxxx.""".stripMargin,
```
systay: The `Foreach` operator executes a nested loop between the LHS and RHS - just like `Apply` it takes a row from the LHS and feeds it to the operator tree on the RHS using the `Argument` operator. It will pull all the results from the RHS, but will just discard them, just yield all the rows coming in from the LHS.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -350,17 +679,47 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+      queryText = """MATCH (s:Person {name: 'me'}) MERGE (s)-[:FRIENDS_WITH]->(s)""",
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("Lock"))
+    )
+  }
+
+  @Test def optional() {
+    profileQuery(
+      title = "Optional",
+      text =
+        """xxx For use in optional match.""".stripMargin,
```
systay: The `Optional` operator is used to solve some `OPTIONAL` queries. It will pull data from it's source, and simply pass it through if any data exists. If nothing is returned by it's source, it will yield a single row with all columns set to `null`.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -441,16 +836,287 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+          |on `:Team(name)` and `:Team(id)`, any node that would be found by the `UniqueIndexSeek`
+          |must be the very same node, or the constraints would be violated.
+        """.stripMargin,
+      queryText =
+        """MERGE (t:Team {name: 'Engineering', id: 42})""".stripMargin,
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("AssertSameNode"))
+    )
+  }
+
+  @Test def nodeHashJoin() {
```
systay, line 1022: Hash Joins General
-=-=-=-=-=-=-=-=-=-
Hash joins have two inputs: the build input and probe input. The query planner assigns these roles so that the smaller of the two inputs is the build input. The build input is pulled in eagerly and used to build a probe table. Once this is finished, the probe table is checked for each row coming from the probe input side.

In query plans, the build input is always the left operator, and the probe input the right operator.

Hash Join Operators:

Node Hash Join:
A variation that does the hash join on node ids, which means it can be done in a faster way, using primitive types and arrays.

Value Hash Join:
Allows for arbitrary values to be used as the join key. Most often used to solve predicates of the form equality predicates between two property columns.

Outer Hash Join:
Instead of discarding rows that are not found in the probe table, will yield a single row with null instead.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -441,16 +836,287 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+          |RETURN other.name""".stripMargin,
+      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("SelectOrSemiApply"))
+    )
+  }
+
+  @Test def selectOrAntiSemiApply() {
+    profileQuery(
+      title = "Select Or Anti Semi Apply",
+      text =
+        """The `SelectOrAntiSemiApply` operator tests for the absence of a pattern predicate and evaluates a predicate.""".stripMargin,
```
systay: The`SelectOrAntiSemiApply` operator is used to evaluate `OR` between a predicate and a negative pattern predicate. If the predicate returns `true`, the pattern is not tested. If the predicate returns `false` or `null`, the operator will test the pattern instead.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -191,66 +297,269 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
       assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("NodeIndexSeek"))
     )
   }
 
+  @Test def nodeByUniqueIndexSeek() {
+    profileQuery(
+      title = "Node Unique Index Seek",
+      text = """The `NodeUniqueIndexSeek` operator finds nodes using an index seek within a unique index. The node variable and the index used is shown in the arguments of the operator.
+               |If the index is not unique, the operator is instead called <<query-plan-node-index-seek, NodeIndexSeek>>.""".stripMargin,
+      queryText = """MATCH (t:Team {name: 'Malmo'}) RETURN t""",
```
systay: If the index seek is used to solve a MERGE clause, it will also be marked with `(Locking)`. This makes it clear that any nodes returned from the index will be locked, in order to protect from concurrent conflicting updates.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: I have changed the text accordingly 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
petraselmer: 👍

petraselmer:
@mariascharin please merge this and fwd-merge accordingly

PR #182, opened 2018-02-14
Add 'mode' to procedure reference

jjaderberg:


mariascharin (commented):
The added `Mode` column looks good but there is something strange going on with the `Signature` column.

It is in black instead of red - which I must say I like much better. However, it looks like the double colon in the signature (`::`) is interpreted as the double colon in a labeled list so that instead of `db.constraints() :: (description :: STRING?)` it becomes `db.constraints() (description :: STRING?)`.

PR #192, opened 2018-03-15
Move QueryPlanTest to DocumentingTest

sherfert:


jjaderberg (changes requested):
This brings the tests onto DocumentingTest but none of the prose. Using DocumentingTest means expressing both tested and untested content through the DSL. Everything in `cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc` and its children.

jjaderberg:
@sherfert A number of changes went into 3.3 and 3.4 that should conflict with this PR, it would probably be easier starting over targeting one of those branches instead (from 3.4 is fine in this case).

craigtaverner:
This work should be re-attempted (ported or rewritten) on a newer branch, like 4.0.

PR #194, opened 2018-03-27
Cypher CALL docs should link to docs for building procs and to built-in proc reference

jjaderberg:
call-link-admon-preview


mariascharin (approved):
I might have shifted the order of these. Here is a proposal. 

Neo4j ships with a number of built-in procedures. 
For a list of these, see <<operations-manual#ref-procedures, Operations Manual -> Built-in procedures>>.

Procedures can also be custom-developed and deployed to the database. 
See <<procedures>> for details.

It is totally fine as-is though, so feel free to ignore if you prefer.

PR #195, opened 2018-03-29
The nouns and descriptions correspond inversely

muei:


petraselmer:
@muei Thanks for noticing this error and fixing it - well-spotted! I wonder if you could please raise this against the `3.1` branch (as the error goes all the way back to that version, and we still need to maintain it)? Many thanks.

muei:
If it need to be internationalized, I can take part in Chinese Translation

mariascharin:
Superseded by #197 .

PR #198, opened 2018-04-04
IN Operator revamp

petraselmer:
* Provided a more detailed explanation of how `IN` works
* Provided a workaround to prevent unexpected results when checking if a node has a label in a list of provided labels


petraselmer:
@mariascharin please fwd merge. This PR seeks to plug a gap in the subtleties of list membership semantics, and provides a query workaround: motivated by a high-priority customer ticket.

PR #210, opened 2018-05-03
Don't lie and claim we support Cypher 3.2

fickludd:


craigtaverner (changes requested):
But where is 3.1? We support that for access to the RULE planner still, I believe.

sherfert:
But not officially @craigtaverner . Just out of technical necessity.

PR #230, opened 2018-07-05
Adapt Cypher docs to changes around internal execution result

fickludd:
Also took the opportunity to copy some test classes from the Neo4j repo
to reduce the coupling between the repos.


jjaderberg (approved):
👍

PR #235, opened 2018-07-25
Adapt to neo4j change getStoreDir() -> databaseDirectory()

burqen:


burqen:
Docstest will break in other ways very soon and all of the issues will be fixed in one batch. Closing this PR.

PR #239, opened 2018-08-01
Execution Plan Section Revamp

petraselmer:


cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
+* <<execution-plan-introduction, Introduction>>
+* <<execution-plan-operators-summary, Execution plan operators>>
+* <<execution-plans-dbhits, DbHits (Database hits)>>
+* <<query-shortestpath-planning, Shortest path planning>>
+
+
+[[execution-plan-introduction]]
+== Introduction
+
 The task of executing a query is decomposed into _operators_, each of which implements a single unit of work.
```
craigtaverner: The phrase 'single unit of work' seems to imply the plan is broken down into the smallest possible _atomic_ units, which is not true. Some operators perform very small indivisible actions, while other perform quite complex operations. I would like to think we could re/word this somehow. Perhaps as simply as:

> The task of executing a query is decomposed into _operators_, each of which implements a specific piece of work.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
-The operators are combined into a structure called an _execution plan_.
+The operators are combined into a tree-like structure called an _execution plan_, in which each operator is represented as a node in the tree.
+Each operator takes as input zero or more rows, and produces as output zero or more rows -- the output from one operator thus forms the input for the next operator.
```
craigtaverner: Since you mention 'tree', perhaps you should define the purpose of the operators at the branch points. How about:

> Operators that join two branches will effectively combine input from two incoming streams and produce a single output.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
+
+.Evaluation model
+The execution plan is evaluated from the bottom up, which means that evaluation begins at the _leaf_ nodes of the tree.
```
craigtaverner: Bottom-up makes sense in the ascii-art version of the plan. But the browsers graph view has it inverted with the leaves at the top. I'm not sure how we should word that?

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
+Leaf nodes have no input rows and generally comprise operators such as scans and seeks which obtain the data directly from the storage engine, thus incurring a <<execution-plans-dbhits, database hit>>.
```
craigtaverner: Should we pluralize dbhits because the count would be the number of rows produced? In fact the code for index seek seems to add one hit for calling the index, followed by one dbhit per row produced. So a seek that returns 100 records would incur 101 dbHits.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
+Any rows produced by leaf nodes are then piped into their parent nodes, which in turn pipe their rows to their parent nodes and so on, all the way up to the root node, which produces the final results of the query.
+
+[[eagerness-laziness]]
+.Eager and lazy evaluation
+In general, query evaluation is _lazy_: most operators start piping their rows as soon as they are produced to their parent operators.
+This means that a child operator may not be fully exhausted before the parent operator starts consuming the input rows produced by the child.
 
+However, some operators, such as those used for aggregation and sorting, need to accumulate _all_ their rows before piping any of them to their parent operator(s); i.e. such an operator needs to complete execution in its entirety before _any_ rows are sent to its parent(s) as input.
+These operators are called _eager_ operators, and are denoted as such in the execution plan.
```
craigtaverner: Are they? Looking at the code I see three eager operators: Eager, Aggregation and Distinct. I also see that Aggregation will display as `EagerAggregation` as you say, but `Distinct` does not. And yet it is still an eager operation, based in fact on the same internal pipe.

craigtaverner: Perhaps just listing these three cases:

> Eager operators are: `Eager`, `EagerAggregation` and `Distinct`

The new work on index-backed order-by might cause previously eager operations to be less eager. For example, a `TopN` operation could become a pure streaming operation on a pre-sorted stream (sorted from the original index, so no intermediate eager sort).

craigtaverner: I've realized that there are more eager pipes:

* Eager
* Aggregation (inlcudes both EagerAggregation and Distinct)
* Sort
* Top
* NodeHashJoin
* OuterHashJoin
* ValueHashJoin

Many of these might become lazy under some specific circumstances due to the 'interesting order' work in the 'index-backed order-by' project.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
+.Statistics
 Each operator is annotated with statistics.
 
 `Rows`::
-The number of rows that the operator produced. Only available if the query was profiled.
+The number of rows that the operator produced.
+This is only available if the query was profiled.
 `EstimatedRows`::
-If Neo4j used the cost-based compiler you will see the estimated number of rows that will be produced by the operator.
+This is the estimated number of rows -- i.e. the approximate number of rows based on the available statistical information -- that will be produced by the operator.
```
craigtaverner: Perhaps _will be_ could be _is expected to be_

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
 The compiler uses this estimate to choose a suitable execution plan.
 `DbHits`::
 Each operator will ask the Neo4j storage engine to do work such as retrieving or updating data.
 A _database hit_ is an abstract unit of this storage engine work.
 The actions triggering a database hit are listed <<execution-plans-dbhits, here>>.
```
craigtaverner: Following this link I see text that claims a single dbhit for each of a long list of actions, while in fact many of those actions generate many dbhits. Perhaps this should be clarified?

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
 
 To produce an efficient plan for a query, Neo4j requires information about the database, such as which indexes and constraints are present (see <<cypher-schema>>).
```
craigtaverner: The wording 'Neo4j requires information about the database' sounds strange to me. Internally is it 'the cypher planner requires information about the kernel stores and indexes', and I wonder if a better public version would be 'the Cypher planner requires information about the Neo4j database'?

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
-Statistical information maintained by the database is also used to generate an efficient execution plan; i.e. the query engine uses this information to determine which access patterns will produce the best execution plan. 
+Statistical information maintained by the database is also used to generate an efficient execution plan; i.e. the query engine uses this information to determine which access patterns will produce the best execution plan.
```
craigtaverner: Here you use the term 'query engine' where above you had the term 'Neo4j' and I had suggested the term 'Cypher planner'. We should pick one term. I like 'Cypher planner', but not you've used less of the term Cypher in many places, and wonder if this is a move towards GQL? I think we should not worry about that before GQL becomes a real thing, at which point we can make real changes.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
 . The number of relationships by type.
+. Selectivity per index.
 . The number of relationships by type, ending with or starting from a node with a specific label.
 
-More information about how the statistics are kept up to date, as well as configuration options about managing query replanning can be found in the <<operations-manual#statistics-execution-plans, Operations Manual -> Statistics and execution plans>>
+More information about how the statistics are kept up to date, as well as configuration options about managing query replanning and caching can be found in the <<operations-manual#statistics-execution-plans, Operations Manual -> Statistics and execution plans>>
 
-See <<how-do-i-profile-a-query>> for how to view the execution plan for a query.
+<<query-tuning>> describes how to tune Cypher queries.
+In particular, see <<how-do-i-profile-a-query>> for how to view the execution plan for a query and <<query-using>> for how to use planner _hints_ to influence the decisions of the planner when building an execution plan for a query.
```
craigtaverner: Do we have a link to a section on these _planner hints_? Satia and I are currently trying to clarify the config setting for `cypher.hints_error`, and there is some subtlety in what is actually a _hint_. We want to clean that up. Notably, the planner hints include in-cypher clauses like `USING INDEX` and `USING JOIN`, but do not include the query prefix `planner=idp`, which is not a hint and must be followed. However, there are runtime hints, the prefix `runtime=compiled` which is just a _hint_ and will not fail unless you set `cypher.hints_error=true`, in which case it does (as do the `USING ...` hints).

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,45 +3,69 @@
...
 
-For a deeper understanding of how each operator works, refer the relevant section.
+For a deeper understanding of how each operator works, refer to the relevant section in <<execution-plan-operators-summary, this table>>.
```
craigtaverner: I'm not a fan of links named _this table_ or _here_ or similar. Can we not name it?

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
craigtaverner: Also I'm not a fan of links named _this table_ or _here_ or similar. Can we not name it?

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -6,17 +6,17 @@
...
-At least data should get filtered out as early as possible in order to reduce the amount of work that has to be done at later stages of query execution.
-This also goes for what gets returned: avoid returning whole nodes and relationships -- instead, pick the data you need and return only that.
+At the very least, data should get filtered out as early as possible in order to reduce the amount of work that has to be done in the later stages of query execution.
+This also applies to what gets returned: returning whole nodes and relationships ought to be avoided in favour of selecting and returning only the data that is needed.
 You should also make sure to set an upper limit on variable length patterns, so they don't cover larger portions of the dataset than needed.
 
-Each Cypher query gets optimized and transformed into an execution plan by the Cypher execution engine.
-To minimize the resources used for this, make sure to use parameters instead of literals when possible.
+Each Cypher query gets optimized and transformed into an <<execution-plan-introduction, execution plan>> by the Cypher execution engine.
```
craigtaverner: Above we have used the terms _Neo4j_, _query planner_ and _query engine_ to mean the same thing as here termed _Cypher execution engine_. Perhaps we should increase consistency?

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -63,7 +63,7 @@ Here we detail the available versions:
 [[cypher-planner]]
 === Cypher planner
 
-Each query is turned into an execution plan by something called the _execution planner_.
+Each query is turned into an <<execution-plan-introduction, execution plan>> by the _execution planner_.
```
craigtaverner: See the previous comment. This line adds a fifth version of the term to the list!

craigtaverner (changes requested):

craigtaverner:
Perhaps the section `Execution plan operators at a glance` should have an extra column on the table called `Eager`? That would clarify better which operator is eager or not.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: Good catch - reworded. 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: I have changed this: the text now refers to the summary table, and I have added a new column to the summary table to make it  easier for readers. 

I think the changes induced by index-backed needs to be done for 3.5 when the work is done.

There is also a card for 3.4 and `Outerhashjoin` being eager - this operator is not present in 3.3

petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: The "query-using" is a hyperlink to the hints section. I think you and Satia should make the changes once this PR has been merged

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -52,89 +77,91 @@ This table comprises all the execution plan operators ordered lexicographically.
...
-| <<query-plan-union, Union>>                                                | Concatenates the results from the right-hand operator with the results from the left-hand operator. | |
-| <<query-plan-unwind, Unwind>>                                              | Returns one row per item in a list. | |
-| <<query-plan-value-hash-join, ValueHashJoin>>                       | Executes a hash join on arbitrary values. | |
-| <<query-plan-varlength-expand-all, VarLengthExpand(All)>>                                   | Traverses variable-length relationships from a given node. | |
-| <<query-plan-varlength-expand-into, VarLengthExpand(Into)>>                                   | Finds all variable-length relationships between two nodes. | |
-| <<query-plan-varlength-expand-pruning, VarLengthExpand(Pruning)>>                                   | Traverses variable-length relationships from a given node and only returns unique end nodes. | |
+|Name                                                                       | Description | Leaf? | Updating? | Eager?
```
craigtaverner: I now have a better idea. Rename the column `Considerations`, and put the word `Eager` in each place you currently have a `Y`. This has two advantages:

* People reading the table way down, don't have to scroll up to see what the `Y` actually means
* We can deal with more subtle situations like partially eager operations (mostly a concern once we implement index-backed order-by, which is 3.5, but it is not wrong to have this forward compatible)

craigtaverner (approved):

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,34 +3,39 @@
...
 == Introduction
 
 The task of executing a query is decomposed into _operators_, each of which implements a specific piece of work.
-The operators are combined into a tree-like structure called an _execution plan_, in which each operator is represented as a node in the tree.
-Each operator takes as input zero or more rows, and produces as output zero or more rows -- the output from one operator thus forms the input for the next operator.
-Operators that join two branches will effectively combine input from two incoming streams and produce a single output.
+The operators are combined into a tree-like structure called an _execution plan_.
+Each operator in the execution plan is represented as a node in the tree.
+Each operator takes as input zero or more rows, and produces as output zero or more rows.
+The output from one operator becomes the input for the next operator.
```
petraselmer: I would like this sentence to be much more closely tied to the previous sentence, as this sentence qualifies the former. (Along the lines of what I originally wrote)

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,34 +3,39 @@
...
+Operators that join two branches in the tree combines input from two incoming streams and produce a single output.
```
petraselmer: "combines" -> "combine" (plural operators)

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,34 +3,39 @@
...
 
 .Evaluation model
-Evaluation of the execution plan begins at the _leaf_ nodes of the tree.
-Leaf nodes have no input rows and generally comprise operators such as scans and seeks which obtain the data directly from the storage engine, thus incurring <<execution-plans-dbhits, database hits>>.
-Any rows produced by leaf nodes are then piped into their parent nodes, which in turn pipe their rows to their parent nodes and so on, all the way up to the root node, which produces the final results of the query.
+Evaluation of the execution plan begins at the leaf nodes of the tree.
+Leaf nodes have no input rows and generally comprise operators such as scans and seeks.
+These operators obtain the data directly from the storage engine, thus incurring <<execution-plans-dbhits, database hits>>.
```
petraselmer: There ought to be no carriage returns for this sentence, as the paragraph as a whole talks about leaf operators. Breaking it up into multiple paragraphs seems odd.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,34 +3,39 @@
...
+The root node produces the final results of the query.
 
 [[eagerness-laziness]]
 .Eager and lazy evaluation
-In general, query evaluation is _lazy_: most operators start piping their rows as soon as they are produced to their parent operators.
+In general, query evaluation is _lazy_: most operators pipe their output rows to their parent operators as soon as they are produced.
 This means that a child operator may not be fully exhausted before the parent operator starts consuming the input rows produced by the child.
 
-However, some operators, such as those used for aggregation and sorting, need to accumulate _all_ their rows before piping any of them to their parent operator(s); i.e. such an operator needs to complete execution in its entirety before _any_ rows are sent to its parent(s) as input.
+However, some operators, such as those used for aggregation and sorting, need to aggregate all their rows before they can produce output.
```
petraselmer: "However, some operators" -> "However some operators" (my original 'bad')

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -3,34 +3,39 @@
...
+Such operators need to complete execution in its entirety before any rows are sent to their parents as input.
```
petraselmer: "Such operators" -> "Such an operator" (rest of sentence assumes singular)

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -42,14 +47,18 @@ The number of rows that the operator produced.
 This is only available if the query was profiled.
 `EstimatedRows`::
 This is the estimated number of rows -- i.e. the approximate number of rows based on the available statistical information -- that is expected to be produced by the operator.
+`EstimatedRows`::
```
petraselmer: I think this is a typo? Identical to line 48

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -42,14 +47,18 @@ The number of rows that the operator produced.
...
+This is the estimated number of rows that is expected to be produced by the operator.
```
petraselmer: Same with this sentence - see line 49

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
```diff
@@ -42,14 +47,18 @@ The number of rows that the operator produced.
...
+The estimate is an approximate number based on the available statistical information.
```
petraselmer: And the same...

Maybe you meant to delete lines 48 - 49?

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -93,31 +93,30 @@ These index decisions are only valid until the schema changes, so adding or remo
 === Cypher runtime
 
 Using the execution plan, the query is executed -- and records returned -- by the Cypher query planner, or _runtime_.
-Depending on whether Neo4j Enterprise Edition or Neo4j Community Edition is used, there are three different runtimes available:
+Depending on whether Neo4j Enterprise Edition or Neo4j Community Edition is used, there are three different runtimes available.
+In Enterprise Edition, the Cypher query planner selects runtime with fallbacks as follows:
```
petraselmer: "selects runtime" -> "selects the runtime"

petraselmer: "runtime with fallbacks as follows" -> "runtime, falling back to alternative runtimes as follows"

petraselmer (commented):
@jjaderberg apparently I cannot request changes, even though that is what I wanted to do :)

Please make the following changes before merging. Thanks!

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: Long sentences are difficult to parse. I have added a "This means that.." to tie them closer without joining them.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: :+1:

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: It is a single paragraph. It might look like there's an empty line in the diff because the changed line is so long it needs to be soft-broken to display it.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: The comma is fine, it helps both with breathing and with eliminating the ambiguity of "however". (I.e., it doesn't mean "in whichever manner" in this context. Compare: "However some operators might behave at a football game, they really are rather a timid bunch if you meet them in a different setting.")

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: Not sure I see it. The "its" refers to "execution". Compare: "All contestants must complete the course in its entirety."

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: Yes! Thanks for catching.

cypher/cypher-docs/src/docs/dev/execution-plans.asciidoc
jjaderberg: :+1:

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
jjaderberg: :+1:

PR #241, opened 2018-08-02
Schema - Indexes and Constraints - revamp

petraselmer:


cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
```diff
@@ -33,10 +33,9 @@ Unique node property constraints can be created, along with node and relationshi
  ** <<constraints-drop-a-node-key, Drop a Node Key>>
  ** <<constraints-create-a-node-that-complies-with-a-node-key, Create a node that complies with a Node Key>>
  ** <<constraints-create-a-node-that-violates-a-node-key, Create a node that violates a Node Key>>
+ ** <<constraints-removing-a-node-key-constrained-property, Removing a `NODE KEY`-constrained property>>
```
craigtaverner, line 36: How does this differ from `Drop a Node Key` above?

cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
```diff
@@ -133,6 +132,6 @@ include::create-a-node-that-complies-with-a-node-key.asciidoc[leveloffset=+1]
 
 include::create-a-node-that-violates-a-node-key.asciidoc[leveloffset=+1]
 
-//include::removing-a-node-key-constrained-property.asciidoc[leveloffset=+1]
+include::removing-a-node-key-constrained-property.asciidoc[leveloffset=+1]
```
craigtaverner, line 135: How does this differ from `Drop a Node Key` above?

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
```diff
@@ -8,24 +8,27 @@ This section explains how to work with indexes in Neo4j and Cypher.
...
+* <<schema-index-substring-search-using-contains-single-property-index, Substring search using `CONTAINS` (single-property index)>>
+* <<schema-index-existence-check-using-exists-single-property-index, Existence check using `exists` (single-property index)>>
 * <<explicit-indexes-procedures, Use built-in procedures to manage and use explicit indexes>>
 
 
 [[query-schema-index-introduction]]
 == Introduction
 
-A database index is a redundant copy of information in the database for the purpose of making retrieving said data more efficient.
+A database index is a redundant copy of the data in the database for the purpose of making more efficient the retrieval of the aforementioned data.
```
craigtaverner: The phrase `the data` seems to imply `all the data`, while the previous term `information` did not imply `all the information`. If we want to make this change, perhaps clarify with `some of the data`.

craigtaverner: Also, it is not necessarily the retrieval of the redundant data copy. For example, we make a property redundant in an index in order to find nodes. Perhaps we should say _for the purpose of making searches of related data more efficient_.

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
```diff
@@ -34,123 +37,41 @@ Cypher enables the creation of indexes on one or more properties for all nodes t
 * An index that is created on more than one property for any given label is called a _composite index_.
   Differences in the usage patterns between composite and single-property indexes are detailed in the examples below.
 
+[NOTE]
+Use <<query-using, hints>> to force Cypher to use specific indexes.
```
craigtaverner: Is the word _force_ correct? The default is to add a warning to the results, and only generate an error if the user sets the `cypher.hint_error` to `true`.

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
```diff
@@ -34,123 +37,41 @@ Cypher enables the creation of indexes on one or more properties for all nodes t
...
+
+
```
craigtaverner: Why the extra newline?

cypher/cypher-docs/src/docs/dev/schema.asciidoc
```diff
@@ -6,24 +6,27 @@
 This section explains how to work with an optional schema in Neo4j in the Cypher query language.
 --
 
-Neo4j 2.0 introduced an optional schema for the graph, based around the concept of labels.
+Neo4j provides an optional schema for the graph, based around the concept of labels.
```
craigtaverner: Many customers consider the labels and relationship types the _schema_ as they are a bit like table names. And then constraints also fit into this thinking, for example property existance constrants are like requiring the table to have certain columns. When we do property type constraints that also fits into schema. But indexes do not.

So we have a clash of mindsets, how customers see _schema_ versus how we internally see it, particularly around the concept of labels, which are very soft in Neo4j and very unlike 'schema'. But I believe in some cases, like the 3.1 security project, we did give some credence to the idea that labels are schema, separating the 'write' user into one that could 'add labels, relationship types and data' from another that could only 'add data'.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
```diff
@@ -304,18 +304,17 @@ class ConstraintsTest extends DocumentingTestBase with SoftReset {
...
-//    execute("CREATE CONSTRAINT ON (n:Person) ASSERT (n.firstname, n.surname) IS NODE KEY")
-//    execute("CREATE (p:Person {firstname: 'John', surname: 'Wood'})")
-//    testFailingQuery[ConstraintValidationException](
-//      title = "Removing a `NODE KEY` constrained property",
-//      text = "Trying to remove the `surname` property from an existing node `Person`, given a `NODE KEY` constraint on `:Person(firstname, surname)`.",
-//      queryText = "MATCH (p:Person {firstname: 'John', surname: 'Wood'}) REMOVE p.surname",
-//      optionalResultExplanation = "In this case the property is not removed."
-//    )
-//  }
+  @Test def break_node_key_constraint_by_removing_property() {
```
craigtaverner, line 307: Is this used anywhere in the docs?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -39,8 +39,8 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
   )
 
   override val properties = Map(
-    "andres" -> Map("firstname" -> "Andres", "surname" -> "Taylor"),
-    "mark" -> Map("firstname" -> "Mark", "surname" -> "Needham")
+    "andres" -> Map("firstname" -> "Andres", "surname" -> "Taylor", "age" -> 40, "country" -> "Sweden"),
```
craigtaverner: I have a spatial docs PR in development which also adds to this, with a normal range query. I used 'highScore' instead of 'age' because I felt like I should not expose personal information like age about real people like Andres and Mark. I then used this for a min<n.prop<max type triple query, and then showed the same this for spatial to demonstrate a bounding-box query.

We should merge our two efforts here.

cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
craigtaverner: OK. I think I found it below. So this is not about removing the NODE KEY, but what happens when you remove a property that is constrained. Makes sense now.

cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
craigtaverner: OK. I think I found it below. So this is not about removing the NODE KEY, but what happens when you remove a property that is constrained. Makes sense now.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -169,10 +241,10 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
     sampleAllIndexesAndWait()
 
     profileQuery(
-      title = "Use index with `STARTS WITH`",
+      title = "Prefix search using `STARTS WITH` (single-property index)",
```
craigtaverner, line 244: Do we need the suffix `(single-property index)` when that is still the most common case and effectively implicit?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -182,13 +254,63 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
...
+  }
+
   @Test def use_index_with_exists_property() {
     // Need to make index preferable in terms of cost
     executePreparationQueries((0 to 250).map { i =>
       "CREATE (:Person)"
     }.toList)
     profileQuery(
-      title = "Use index when checking for the existence of a property",
+      title = "Existence check using `exists` (single-property index)",
```
craigtaverner, line 313: I thought the old title read much better.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/WhereTest.scala
```diff
@@ -46,9 +46,9 @@ class WhereTest extends DocumentingTest {
         | ** <<filter-on-dynamic-property, Filter on dynamically-computed property>>
         | ** <<property-existence-checking, Property existence checking>>
         |* <<query-where-string, String matching>>
-        | ** <<match-string-start, Match the beginning of a string>>
-        | ** <<match-string-end, Match the ending of a string>>
-        | ** <<match-string-contains, Match anywhere within a string>>
+        | ** <<match-string-start, Prefix search using `STARTS WITH`>>
```
craigtaverner: I thought the older text read better, and also included the type requirement (String). The new one is more subtle in the type requirement, since `prefix` implies `string`, but not as explicit about it. If people search the docs are they more like to search for `prefix` or `beginning of`?

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/ql/constraints/index.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
petraselmer: Please check now: this seems to be standard, but have altered wording so it is hopefully more nuanced

petraselmer: 👍

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
petraselmer: 👍

cypher/cypher-docs/src/docs/dev/schema.asciidoc
petraselmer: Not sure what is needed here? Please clarify. I, for one, disliked the sentence from the get-go and after your comments, have elected to remove it altogether.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
petraselmer: Yes, see in Constraints section: "Removing a `NODE KEY`-constrained property"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
petraselmer: I think age and co are good properties to use from a pedagogical point of view. However, we should definitely get better graph examples! This is something I've raised with the Docs Team - it needs to be done across the board.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
petraselmer: I think so - we need some mechanism to make this very explicit and clear to the reader. Maybe a table or sub-TOC would be better - but not with the current awful levels of nesting. Perhaps there is something that can be done in the new Cypher-only manual. Leaving it to Docs Team...

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
petraselmer: I think the old title took too long to actually explain what was happening. Here, one sees from the outset what one can do - it's implicit that the "index is being used"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/WhereTest.scala
petraselmer: I've compromised: you're absolutely right in that the string-ness was lost. Have reinstated that part. Have left prefix, suffix etc (as they're more in line with existing ways of referring to this and more technical/terse), as even if the user searches for "beginning of", they will find it as it is mentioned thus in the description.

petraselmer: 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
craigtaverner: How do we deal with this case then? Invent new anonymous names? Get permission from Mark and Andres to publish their ages?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
craigtaverner: Then my only remaining complaint is the `(single-property index)`. But, as you said before, the Docs team could perhaps take a decision on that.

craigtaverner (approved):

petraselmer:
This is now ready to merge, @jjaderberg

craigtaverner:
Due to conflicts with the spatial PR merged into 3.4, we've made a new PR for this at https://github.com/neo4j/neo4j-documentation/pull/244

So when merging #241, null forward merge `git merge origin/3.3 -s ours ...`, and then merge this new PR #244 into 3.4 and merge forward to 3.5.

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
```diff
@@ -34,123 +37,40 @@ Cypher enables the creation of indexes on one or more properties for all nodes t
...
+<<query-using>> describes how to make the Cypher planner use specific indexes (especially in cases where the planner would not necessarily have used them).
+
 Once an index has been created, it will automatically be managed and kept up to date by the database when the graph is changed.
 Neo4j will automatically pick up and start using the index once it has been created and brought online.
 
 include::create-a-single-property-index.asciidoc[]
 
-include::get-a-list-of-all-indexes-in-the-database.asciidoc[]
-
-[[create-a-composite-index]]
```
jjaderberg, line 44: `create-a-composite-index` still referened in _src/docs/dev/ql/constraints/index.asciidoc_

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
jjaderberg: Agree about making it explicit, agree that the suffix is distracting. Let's find another way, but not in this PR.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -182,13 +254,63 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
...
+    }
+
+    sampleAllIndexesAndWait()
+
+    profileQuery(
+      title = "Suffix search using `ENDS WITH` (single-property index)",
+      text =
+        "The `ENDS WITH` predicate on `person.firstname` in the following query will use the `Person(firstname)` index, if it exists. " +
+          "All values stored in the `Person(firstname)` index will be searched, and entries ending with `'hn'` will be returned. " +
+          "This means that although the search will not be optimized to the extent of queries using `=`, `IN`, `>`, `<` or `STARTS WITH`, it is still faster than not using an index in the first place. " +
```
jjaderberg, line 271: Where can I learn about how queries using "`=`, `IN`, `>`, `<`, or `STARTS WITH` are optimized but queries using `ENDS WITH` are not?

jjaderberg (changes requested):
The broken ID blocks. There are some style questions but I think we would be better off revisiting later. `SchemaIndexTest` still needs to be ported to the less old tooling, so we will have occasion to revisit soon.

jjaderberg (approved):

PR #249, opened 2018-08-22
Fix Cypher result serialization bug

jjaderberg:
* can serialize JavaListWrapper, Arrays, etc.
* avoid using internal types
* force being explicit on how to represent each type

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/CypherSerializer.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/CypherSerializer.scala
@@ -22,15 +22,13 @@ package org.neo4j.cypher.internal.compatibility.v3_4.runtime
 import java.time._
 import java.time.temporal.TemporalAmount
 
-import org.neo4j.cypher.internal.runtime.{JavaListWrapper, QueryContext}
+import org.neo4j.cypher.internal.runtime.QueryContext
 import org.neo4j.graphdb.{Node, Path, PropertyContainer, Relationship}
 import org.neo4j.graphdb.spatial.Point
 
 import scala.collection.Map
 
 trait CypherSerializer {
-
-  import scala.collection.JavaConverters._
 
   /*
   Be explicit to force the decision how to represent each type.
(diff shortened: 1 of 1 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/CypherSerializer.scala
```diff
@@ -19,27 +19,46 @@
...
+    case x: TemporalAmount        => x.toString
+    case x: LocalDate             => x.toString
+    case x: LocalDateTime         => x.toString
+    case x: LocalTime             => x.toString
+    case x: OffsetTime            => x.toString
+    case x: ZonedDateTime         => x.toString
+    case x: Point                 => x.toString
+    case null                     => "<null>"
+    case Some(x)                  => serialize(x, qtx)
+    case x                        => throw new IllegalArgumentException(s"Type ${x.getClass} must be explcitly handled.")
```
fickludd: `explcitly`

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/CypherSerializer.scala
```diff
@@ -19,27 +19,46 @@
...
-    case x: String       => "\"" + x + "\""
-    case v: KeyToken     => v.name
-    case Some(x)         => x.toString
-    case null            => "<null>"
-    case x               => x.toString
+    case x: Node                  => x.toString + serializeProperties(x, qtx)
+    case x: Relationship          => ":" + x.getType.name() + "[" + x.getId + "]" + serializeProperties(x, qtx)
+    case x: Path                  => x.toString
+    case x: Map[_, _]             => makeString(x.asInstanceOf[Map[String, Any]], qtx)
+    case x: java.util.Map[_, _]   => makeString(x.asInstanceOf[java.util.Map[String, Any]].asScala, qtx)
```
fickludd: It is very strange that we expect to get both `scala` and `java` collections here... If we are consuming over any public surface we should only be getting `java` collections, and if we are operating with some internal `ExecutionResult` hackery I would say we should only expect `scala` collections. But accepting both is strange.

fickludd (approved):
I like the improved formatting and readability, as well as not accepting all input. I don't think there is anything critically wrong here, but noted two things that can be addressed or not.

PR #253, opened 2018-09-06
Hello Cypher Manual

jjaderberg:
Can build Cypher Manual outputs close to the Cypher docs tests and static sources.


mariascharin:
@jjaderberg Can you recall the status of this PR? Is it worth pursuing?

jjaderberg:
The PR adds a docs build to produce the Cypher manual html and pdf directly from this repo. Back then we thought that it made sense and was valuable to break out the output-producing build for Cypher docs and put it next to the Cypher doc tests, as it would make it easier for Cypher devs to work with their docs and would better mirror what we do with most newer docs builds. I don't know if that's still considered valuable or important.
As for status.. it worked back then and was mostly (possibly entirely) feature complete. Some things are probably a bit outdated, but in order to say how close to ready it is I would have to review it.

PR #263, opened 2018-09-25
Add Cypher deprecations

fickludd:
Add deprecation warnings to

 * rule planner
 * compiled runtime
 * extract, filter, all, any, none, single

I noticed that these things are also in the refcard, but am unsure of how to deprecate content there, or if it's actually better to just remove deprecated things from there and replace them with the relevant list-comprehension. Which conveniently bring me to wonder: is there any place where list-comprehensions (e.g. `[x IN xs WHERE x.name = "johan" | x.age / 10]`) are presented. I couldn't really find it, maybe we should add more on that.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -112,6 +112,7 @@ An example of this is `CALL db.index.explicit.searchNodes('my_index','email:me*'
 [options="header"]
 |===
 | Feature          | Type | Change | Details
+| `CYPHER runtime=compiled` (Compiled runtime)    | Functionality | Deprecated | The compiled runtime will be discontinued on the next major release. It might still be used for default queries in order to not cause regressions, but explicitly requesting it will not be possible.
 |===
 
 [[cypher-compatibility]]
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -112,7 +112,9 @@ An example of this is `CALL db.index.explicit.searchNodes('my_index','email:me*'
 [options="header"]
 |===
 | Feature          | Type | Change | Details
-| `CYPHER runtime=compiled` (Compiled runtime)    | Functionality | Deprecated | The compiled runtime will be discontinued on the next major release. It might still be used for default queries in order to not cause regressions, but explicitly requesting it will not be possible.
+| `CYPHER runtime=compiled` (Compiled runtime)    | Functionality | Deprecated | The compiled runtime will be discontinued in the next major release. It might still be used for default queries in order to not cause regressions, but explicitly requesting it will not be possible.
+| `extract()`   | Function  | Deprecated | Replaced by <<cypher-list-comprehension, list comprehension>>
+| `filter()`   | Function  | Deprecated | Replaced by <<cypher-list-comprehension, list comprehension>>
 |===
 
 [[cypher-compatibility]]
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ListFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ListFunctionsTest.scala
@@ -44,6 +44,7 @@ class ListFunctionsTest extends DocumentingTest {
     p("Further details and examples of lists may be found in <<cypher-lists>> and <<query-operators-list>>.")
     note {
       p("The function `rels()` has been superseded by `relationships()`, and will be removed in a future release.")
+      p("The functions `extract()` and `filter()` have been deprecated and will be removed in a future release. Consider using a list comprehension (e.g. `[x IN xs WHERE predicate | extraction]`) instead.")
     }
     p(
       """Functions:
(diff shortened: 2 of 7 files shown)


cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -112,6 +112,7 @@ An example of this is `CALL db.index.explicit.searchNodes('my_index','email:me*'
 [options="header"]
 |===
 | Feature          | Type | Change | Details
+| `CYPHER runtime=compiled` (Compiled runtime)    | Functionality | Deprecated | The compiled runtime will be discontinued on the next major release. It might still be used for default queries in order to not cause regressions, but explicitly requesting it will not be possible.
```
petraselmer: "discontinued on the" -> "discontinued in the"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -52,6 +52,9 @@ class PredicateFunctionsTest extends DocumentingTest {
         |* <<functions-exists,exists()>>
         |* <<functions-none,none()>>
         |* <<functions-single,single()>>""")
+    note {
+      p("The functions `all()`, `any()`, `none()` and `single()` have been deprecated and will be removed in a future release. Consider using a list-comprehension (e.g. `[x IN xs WHERE predicate | extraction]`) together with `size()` instead.")
```
petraselmer: "using a list-comprehension" -> "using a list comprehension"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ListFunctionsTest.scala
```diff
@@ -44,6 +44,7 @@ class ListFunctionsTest extends DocumentingTest {
     p("Further details and examples of lists may be found in <<cypher-lists>> and <<query-operators-list>>.")
     note {
       p("The function `rels()` has been superseded by `relationships()`, and will be removed in a future release.")
+      p("The functions `extract()` and `filter()` have been deprecated and will be removed in a future release. Consider using a list comprehension (e.g. `[x IN xs WHERE predicate | extraction]`) instead.")
```
petraselmer: I wonder if it would be a good idea to add `[role=deprecated]` (as you did above, for the rule planner in query tuning) to the detailed text for `extract` and `filter`? People may jump to the details immediately without going from the top of the page, and it doesn't hurt to really, really drive home that these functions are on their way out. 

I would also add, to the function detail section - your second sentence "Consider using a list com..." so that people know how to achieve the same functionality

petraselmer: I also would link "list comprehension" to the relevant section in the manual. It's not much - i.e. it is far from perfect or even good - but at least it provides context and extra reading for those users who haven't used list comprehensions. 

The way to do this is to replace "list comprehension" with "<<cypher-list-comprehension, list comprehension>>"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
petraselmer: I wonder if it would be a good idea to add [role=deprecated] (as you did above, for the rule planner in query tuning) to the detailed text for all, any etc? People may jump to the details immediately without going from the top of the page, and it doesn't hurt to really, really drive home that these functions are on their way out.

I would also add, to the function detail section - your second sentence "Consider using a list com..." so that people know how to achieve the same functionality

petraselmer: I also would link "list comprehension" to the relevant section in the manual. It's not much - i.e. it is far from perfect or even good - but at least it provides context and extra reading for those users who haven't used list comprehensions.

The way to do this is to replace "list comprehension" with "<<cypher-list-comprehension, list comprehension>>"

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ListFunctionsTest.scala
petraselmer: Also, can you give an actual example using a list comprehension that achieves the same effect? Probably sufficient to do it for just one of `extract` and `filter`; i.e. as long as there is something on the page readers can see.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
petraselmer: Also, can you give an actual example using a list comprehension that achieves the same effect? Probably sufficient to do it for just one of the 4 functions; i.e. as long as there is something on the page readers can see.

petraselmer (changes requested):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
mariascharin: @fickludd I propose removing deprecated stuff from the refcard. It is the Cypher Manual that is the comprehensive and authoritative reference.

fickludd: Thank you both! For the list-predicates (all, any, single and none) we have actually swung around to not wanting to deprecate these anymore, so changing this PR to reflect that.

petraselmer (approved):

PR #264, opened 2018-09-27
Prettify lines with line breaks

sherfert:
Part 1: Make Prettifier configurable to keep line breaks. Make thart available through `preformattedQuery`

Part 2: Use the `preformattedQuery` where appropriate in the Temporal docs.


fickludd:
Moving the prettifier from Neo4j to docs is clearly an improvement. I guess the other place where it could make sense is actually in the front-end, because it could be useful to anyone who needs the parser. Let's remember that in case CAPS need something similar.

PR #269, opened 2018-10-02
Add ee tag to ee-only constraints

mariascharin:
Node Key is introduced in 3.2 - Must add tags to that content on forward-merge!


davidoliverSP2 (approved):
LGTM! 👍

PR #276, opened 2018-10-23
Mention limitations of Lucene index

mariascharin:


mariascharin:
In 3.3, add:

```
In a composite index, this limit is applicable to each individual property. 
This means that a composite index key can hold values that together are larger than 32766 bytes but no single value can be larger.
```

In 3.4 and 3.5, change to reference to Ops Man.

PR #279, opened 2018-11-01
Add composite index examples to the Cypher refcard

Lojjs:
The Cypher refcard didn't mention composite indexes anywhere, now it does.

### DIFF
--- a/cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/IndexTest.scala
+++ b/cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/IndexTest.scala
@@ -44,13 +44,15 @@ class IndexTest extends RefcardTest with QueryStatisticsTestSupport {
   }
 
   override val properties: Map[String, Map[String, Any]] = Map(
-    "A" -> Map("name" -> "Alice"),
-    "B" -> Map("name" -> "Timothy"))
+    "A" -> Map("name" -> "Alice", "age" -> 18 ),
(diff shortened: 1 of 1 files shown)


cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/IndexTest.scala
```diff
@@ -86,6 +96,18 @@ RETURN n
 
 An index can automatically be used for the `IN` list checks.
 
+###assertion=match parameters=nameandage
+//
+
+MATCH (n:Person)
+WHERE n.name = $value and n.age = $value2
+RETURN n
```
craigtaverner, line 105: Looks like there is a missing newline before the RETURN. This will probably not be rendered the same as the others above.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/IndexTest.scala
```diff
@@ -86,6 +96,18 @@ RETURN n
...
+###
+
+A composite index can be automatically used for equality comparison of both properties.
+Note that a predicate on only one of the properties
```
craigtaverner: Perhaps more general: `Note that there needs to be predicates on all properties of the composite index for it to be used.`

craigtaverner (changes requested):

craigtaverner (approved):

PR #281, opened 2018-12-03
Describe limitations of composite indexes better

Lojjs:
Also clarify that you cannot fallback from a composite index to a single one like in SQL

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
@@ -71,7 +71,7 @@ Queries containing the following types of predicates on properties in the index 
 * substring search: `CONTAINS`
 
 Composite indexes require predicates on all properties indexed.
-If there are predicates on only a subset of the indexed properties, the Cypher will not be able to use the composite index.
+If there are predicates on only a subset of the indexed properties, it will not be possible to use the composite index.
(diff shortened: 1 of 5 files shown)


craigtaverner (changes requested):
The complete rendered docs do not include the new content. I think the section id `schema-index-single-vs-composite-index` need to be included somewhere to make this active.

craigtaverner (approved):

cypher/cypher-docs/src/docs/dev/ql/schema-index/index.asciidoc
```diff
@@ -58,52 +58,58 @@ Neo4j will automatically pick up and start using the index once it has been crea
...
-include::get-a-list-of-all-indexes-in-the-database.asciidoc[]
+* existence check: `exists(n.prop)`
+* range search: `n.prop > value`
+* prefix search: `STARTS WITH`
+* suffix search: `ENDS WITH`
+* substring search: `CONTAINS`
 
-include::drop-a-single-property-index.asciidoc[]
+Composite indexes require predicates on all properties indexed.
+If there are predicates on only a subset of the indexed properties, the Cypher will not be able to use the composite index.
```
Lojjs: This seems clearer, but you should remove `the` in `the Cypher`

Lojjs (commented):
I found an error in your text, otherwise it looks fine (apart from that I cannot see the rendering...)

PR #282, opened 2018-12-04
Use toInteger over deprecated toInt

jjaderberg:


Lojjs:
This was already fixed in #283

PR #286, opened 2018-12-10
Force bolt connector settings to be recognized

jjaderberg:
Some bolt settings are not recognized by default.
By explicitly adding them while building the config we can force them to
be included.
This is going to break.
But it gives desirable results for now.

The config settings framework is very difficult to use for some things.
Fingers crossed it will be improved between this fix, and when this fix
breaks.


mariascharin:
Note to self: Build and check that dynamic settings page is still OK with this change.

PR #294, opened 2019-01-08
Remove dead link in Cypher Man

mariascharin:


mariascharin:
This PR removes a reference to a section that was previously in the Developer Manual, but is now in the _Getting Started Guide_. The Getting Started Guide is less formal than the Cypher Manual, with more casual language etc., so I don't want to keep the reference. If more examples are needed, they should be added to the Cypher Manual itself.

PR #309, opened 2019-02-19
HTTP Console removal

PetrJanouch:


zhenlineo:
@mariascharin We would like to remove this test in docs to prepare for removal of HTTP console from the neo4j on the server side. The HTTP console endpoint is not documented and this test seems not testing anything.

PR #310, opened 2019-02-20
Scala 2.12 compatibility

Lojjs:


jjaderberg:
@Lojjs You're not actually changing the `scala.binary.version`, right? It's still 2.11

jjaderberg (approved):
Parboiled dependency needs a version. Otherwise LGTM.

PR #322, opened 2019-03-20
4.0 improved functions

craigtaverner:
Restructuring the cypher functions page for improved symmetry


craigtaverner:
Since this bug is in older docs, we replace this PR with https://github.com/neo4j/neo4j-documentation/pull/326

PR #324, opened 2019-03-21
Fix chapter user-defined aggregation functions

mariascharin:


mariascharin:
Connected to https://github.com/neo-technology/neo4j-manual-modeling/pull/755

craigtaverner (changes requested):
This seems to clash with the work I'm doing here: https://github.com/neo4j/neo4j-documentation/pull/322

My work is, I believe, a super-set of what you have done. I will be porting it back to 3.4 and 3.5 (or perhaps 3.5 only if that is preferable) now.

PR #351, opened 2019-05-23
4.0 start of cleaning

MishaDemianenko:


kernel/src/test/java/examples/DeadlockDocTest.java
```diff
@@ -22,57 +22,51 @@
...
     }
 
     @Test
-    public void transactionWithTemplate() throws InterruptedException
+    void transactionWithTemplate( GraphDatabaseService databaseService )
     {
-        GraphDatabaseService graphDatabaseService = rule.getGraphDatabaseAPI();
-
         // tag::template[]
         TransactionTemplate template = new TransactionTemplate(  ).retries( 5 ).backoff( 3, TimeUnit.SECONDS );
```
klaren, line 50: This was moved to `*.internal.helpers.*`. Let's make a note to add it back or remove this snippet altogether.

klaren (approved):
Good start!

kernel/src/test/java/examples/DeadlockDocTest.java
MishaDemianenko: yep yep, ideally when we will enable our super check that will check that this repo uses only public api we will remove everything that is not part of that :) (in the bright future)

PR #353, opened 2019-05-28
Cleanup custom docs test database factories.

MishaDemianenko:
Remove all custom test database rules and supporting classes.
Use only database builders and extensions to build databases that are
part of public API.


cypher/graphgist/src/main/java/org/neo4j/doc/cypherdoc/CypherDoc.java
```diff
@@ -74,14 +74,12 @@ public static String parse( String input, File parentDirectory, String url )
     {
         List<Block> blocks = parseBlocks( input );
 
-        EphemeralFileSystemAbstraction fs = new EphemeralFileSystemAbstraction();
         //TODO remove config when compiled plans are feature complete
```
klaren, line 77: :hourglass_flowing_sand:

server-docs/src/test/java/org/neo4j/doc/server/ExclusiveServerTestBase.java
```diff
@@ -19,33 +19,24 @@
...
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.rules.TestName;
 
-import org.neo4j.doc.test.rule.SuppressOutput;
 import org.neo4j.doc.test.rule.TestDirectory;
 
-import static org.neo4j.doc.test.rule.SuppressOutput.suppressAll;
 
 public class ExclusiveServerTestBase
```
klaren, line 29: This group of classes is still duplicated. Can we do something about this?

klaren (approved):

server-docs/src/test/java/org/neo4j/doc/server/ExclusiveServerTestBase.java
MishaDemianenko: aha, those classes are also copy-pasted( and they are waiting for their cleaning PR

PR #367, opened 2019-07-15
Added a section about the valid geographic coordinates

Hunterness:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SpatialTest.scala
```diff
@@ -184,6 +184,13 @@ class SpatialTest extends DocumentingTest {
           })) {
           resultTable()
         }
+        p(
+          """
+            |For the geographic values, there is to note that the `latitude` value should always lie in the interval `[-90, 90]` and any other value
```
craigtaverner: The phrase `there is to note` is not correct English. I suggest changing it to `it is important to note`.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SpatialTest.scala
```diff
@@ -184,6 +184,13 @@ class SpatialTest extends DocumentingTest {
...
+            |outside this range will throw an exception. The `longitude` value should always lie in the interval `[-180, 180]` and any other value
+            |outside this range will be wrapped around to fit in this range.
```
craigtaverner: It might be worth adding that :
> The value for `height` and any Cartesian coordinates are not explicitly restricted, and any value within the allowed range of the signed 64-bit floating point type will be accepted.

craigtaverner (changes requested):

craigtaverner (approved):

PR #374, opened 2019-07-30
4.0 remove stray stuff

mariascharin:
@craigtaverner I took a stab at this (see my Slack message earlier today). Please review and let me know what you think. Thanks in advance!

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/schema-index/fulltext-schema-index.adoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/schema-index/fulltext-schema-index.adoc
@@ -43,13 +43,6 @@ While a composite index applies only to entities that match the indexed label an
 
 For information on how to configure full-text schema indexes, refer to <<operations-manual#index-configuration-fulltext-search,  Operations Manual -> Indexes to support full-text search>>.
 
-[NOTE]
---
-Full-text schema indexes replace the _explicit_ indexes, which are deprecated and will be discontinued in the next major release.
-It is therefore recommended migrate to full-text schema indexes.
-A full description of the differences between full-text schema indexes and explicit indexes is available in <<operations-manual#index-configuration-fulltext-search-deprecation-of-explicit, Operations Manual -> Deprecation of explicit indexes>>.
---
-
 
 [[schema-index-fulltext-search-manage]]
 == Procedures to manage full-text schema indexes
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallTest.scala
@@ -74,14 +74,8 @@ class CallTest extends DocumentingTest {
             |For a list of these, see <<operations-manual#ref-procedures, Operations Manual -> Built-in procedures>>.""")
         p(
           """Users can also develop custom procedures and deploy to the database.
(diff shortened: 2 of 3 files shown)


cypher/cypher-docs/src/docs/dev/ql/schema-index/fulltext-schema-index.adoc
```diff
@@ -43,13 +43,6 @@ While a composite index applies only to entities that match the indexed label an
 
 For information on how to configure full-text schema indexes, refer to <<operations-manual#index-configuration-fulltext-search,  Operations Manual -> Indexes to support full-text search>>.
 
-[NOTE]
```
craigtaverner, line 46: I would think that users of 4.0 would still need to know how to port from 3.x to 4.0. Is there another place in the documentation that covers this? I think it is good to remove these bits and pieces here and there if there is a section of porting from 3.x, but not good to remove this if no such content exists.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallTest.scala
```diff
@@ -110,33 +104,33 @@ class CallTest extends DocumentingTest {
     }
 
     section("Call a procedure with literal arguments", "call-call-a-procedure-with-literal-arguments") {
-      p("""This calls the example procedure `org.neo4j.procedure.example.addNodeToIndex` using literal arguments.
+      p("""This calls the example procedure `dbms.security.createUser` using literal arguments.
```
craigtaverner, line 107: The security procedures are deprecated in 4.0 (and in fact there was even a plan for complete removal, but it seems more likely we'll just deprecate). We should use some other examples.

craigtaverner: We need an example that works on a normal database. Security procedures will only work on system databases (if they are not removed entirely). I favor having a user-built procedure, although a very small and simple one.

craigtaverner: I have another idea. What about the `db.stats.*` procedures? They are new, work on normal databases (not system) and useful to normal users for support requests?

craigtaverner (changes requested):

cypher/cypher-docs/src/docs/dev/ql/schema-index/fulltext-schema-index.adoc
mariascharin: That is a good point, and we should make sure there is a section about this in whatever "migrating to this new and shiny 4.0 release" document that will have to accompany the release. I have made a note of it in a card on our wall.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallTest.scala
mariascharin: @craigtaverner Thank you! I will try it out.

mariascharin: @craigtaverner As agreed in discussions yesterday, I will merge this to get the bug out, and raise a new card to find a different example.

davidoliverSP2 (approved):
Thanks @mariascharin 
I got this to build after rebasing.

PR #378, opened 2019-07-30
Update for the change in config giving type to setting values

Roxling:


mariascharin:
@Roxling Feel free to merge this yourself if it builds cleanly.

PR #379, opened 2019-08-01
Escape curly braces in asciidoc to avoid complaint at compile-time

mariascharin:


davidoliverSP2 (approved):
Works for me, thanks @mariascharin 
👍

PR #385, opened 2019-08-19
changed double to float

AlexicaWright:


davidoliverSP2 (approved):
👍 

@AlexicaWright have you checked with someone in the Cypher team that they concur?

PR #395, opened 2019-09-17
Import com.neo4j.enterprise.edition.factory.EnterpriseDatabaseManagem…

mariascharin:
…entServiceBuilder where needed


mariascharin:
@klaren This is related to what we slacked about earlier today. 
Could you please verify that what I did is correct? 
Thank you in advance!

klaren:
@mariascharin looks correct!

PR #410, opened 2019-10-02
Do not use `WITH` in reltype.

Lojjs:
That did confuse the docs formatter making it line break before the WITH.

I could not see any graph visualizations when building docs locally, so please check that extra careful.


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/MatchTest.scala
```diff
@@ -180,18 +180,17 @@ class MatchTest extends DocumentingTest {
       section("Relationship types with uncommon characters", "rel-types-with-uncommon-chars") {
         val initQuery =
           """MATCH (charlie:Person {name: 'Charlie Sheen'}), (rob:Person {name: 'Rob Reiner'})
-            |CREATE (rob)-[:`TYPE
-            |WITH SPACE`]->(charlie)"""
```
Hunterness, line 184: I know @SaschaPeukert changed this as well when fixing the Windows build, but differently. I don't know what the correct behavior/query should be.

PR #411, opened 2019-10-02
A temporal function of null will return null.

Lojjs:
Maybe this should rather be in 3.4, since the behaviour in the code is changed from 3.4 and forward.


Hunterness (approved):
If the code changed in 3.4 I agree that it would be nice to backport this as well :)

mariascharin:
Superseded by #416

PR #419, opened 2019-10-08
Fix bug in merge example

mariascharin:


mariascharin:
Already fixed in 3.5 and onwards. Merge and null-forward merge!

PR #423, opened 2019-10-10
Removing repository declarations for m2.neo4j.org

lassewesth:
These declarations at best make m2.neo4j.org a SPOF when building docs. Instead, let Maven Central and it's CDN provide released dependencies.

The problem you have been seeing occur whenever Maven sees a dependency declared as a snapshot version. Because it then picks the abandoned m2.neo4j.org to resolve it, fails, and _reports m2.neo4j.org as the failure_ which is just hella misleading.

This change solves all the above mentioned problems once and or all 🎉 

This can be forward merged to 3.2 and 3.3 which have similar repository declarations.

### DIFF
--- a/backup/pom.xml
+++ b/backup/pom.xml
@@ -5,14 +5,14 @@
     <parent>
         <artifactId>neo4j-manual-parent</artifactId>
         <groupId>org.neo4j.doc</groupId>
-        <version>3.1.10-SNAPSHOT</version>
+        <version>3.1.9</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>org.neo4j.doc</groupId>
     <artifactId>neo4j-backup-docs</artifactId>
-    <version>3.1.10-SNAPSHOT</version>
+    <version>3.1.9</version>
 
     <name>Neo4j - Backup Docs</name>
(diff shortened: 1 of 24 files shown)


jjaderberg (changes requested):
Needs to resolve Neo4j parent from somewhere: https://github.com/neo4j/neo4j-documentation/blob/3.1/pom.xml#L4-L8

Who needs to build this for 3.1..3.3 anyway? In 3.4 parent is gone.

If this really needs to work maybe freeze the version to something released?

jjaderberg (approved):

jjaderberg:
If it's acceptable to freeze the Neo4j dependency version, this should work.

lassewesth:
Who's the arbiter?

jjaderberg:
@mariascharin @davidoliverSP2 @AlexicaWright I think this is a good change. Is it ok with you if I merge this?

mariascharin:
@jjaderberg You are welcome to it if you have time :) Thank you!

mariascharin:
@lassewesth I am replacing this one with #520 since we have stopped building any docs previous to 3.4. With that, there shouldn't be anything remaining from Docs point of view that depends on m2.neo4j.org.

mariascharin:
Superseded by #520

PR #431, opened 2019-10-21
Add some more profiling info

pontusmelke:


sherfert (commented):
The first commit looks good. The other commits seem to be part of a different changeset?

PR #435, opened 2019-10-23
Add documentation for subqueries

pontusmelke:
This includes:
- Fixes to allow preformatting queries
- New section on `CALL` subquery
- Update refcard with existential subqueries and `CALL` subquery
- Move section on existential subqueries to `WHERE`


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ExistentialSubQueriesTest.scala
```diff
@@ -34,7 +34,7 @@ class ExistentialSubQueriesTest extends DocumentingTest {
 
     p(
       """An existential subquery can be used to find out if a specified pattern exists at least once in the data.
-         |It can be used in the same way as a predicate expression but it allows you to use `MATCH` and `WHERE` clauses internally.
+         |It can be used in the same way as a path pattern but it allows you to use `MATCH` and `WHERE` clauses internally.
```
SaschaPeukert: That feels a bit like I am allowed to write 

`MATCH EXISTS { ... } RETURN ...`

PR #437, opened 2019-10-29
Added information about named indexes and constraints, the new create index syntax and the deprecation of the old drop syntax

Hunterness:


Hunterness (commented):
Things I want to fix:

- [x] Glossary of keywords can have name as optional
- [x] Any index can be dropped by name
- [x] Named constraints is the wrong type of section (maybe also drop named constraints but it might be correct)
- [x] Missing space in sentence about dropping named constraint
- [x] don't have constraint name that is a keyword (gets all caps)
- [x] Link to `Create a relationship property existence constraint` is broken
- [x] Add something about the name for the execution plan operators
- [ ] ~Switch so we say `This is deprecated` after the sentences about dropping constraints/indexes~
- [x] Add `deprecated` square to the relevant sections instead of writing it out in the text everywhere (see https://neo4j.com/docs/cypher-manual/3.5/schema/index/#explicit-indexes-procedures)

Also wants too see how the refcard looks since I have made some changes there and not yet looked at it.

There is also some stupid formatting (breaking the new index and constraint syntax on the `ON` keyword since there is (at least) a name between the `CONSTRAINT` and `ON`)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -143,8 +147,12 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
         """The `CreateNodePropertyExistenceConstraint` operator creates an existence constraint on a property for all nodes having a certain label.
           |This will only appear in Enterprise Edition.
         """.stripMargin,
-      queryText = """CREATE CONSTRAINT ON (p:Person) ASSERT exists(p.name)""",
-      assertions = (p) => assertThat(p.executionPlanDescription().toString, containsString("CreateNodePropertyExistenceConstraint"))
+      queryText = """CREATE CONSTRAINT existence ON (p:Person) ASSERT exists(p.name)""",
```
Lojjs, line 150: In the above example you write out in the text that `uniqueness` is the name of the constraint. If only reading this example I think it can be a bit confusing whether `existence` is a required keyword since it is so similar `exists` or whether it is a name. Either make it clear in the text as above or maybe name it differently to something that is more clearly a name like `constraint_name`.

The same goes for all other indexes/constraints with names like `existence` and `uniqueness`

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
```diff
@@ -288,7 +334,7 @@ class ConstraintsTest extends DocumentingTestBase with SoftReset {
     generateConsole = false
 
     prepareAndTestQuery(
-      title = "Create a node that complies with a Node Key",
+      title = "Create a node that complies with node key constraints",
```
Lojjs, line 337: I like this change, but for consistency I think you should change it in the remaining places in the constraints chapter as well, e.g. in 5.3.1

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/QueryPlanTest.scala
```diff
@@ -215,7 +231,25 @@ class QueryPlanTest extends DocumentingTestBase with SoftReset {
...
+    )
+  }
+
+  @Test def dropNamedConstraint() {
+    executePreparationQueries {
+      List("CREATE CONSTRAINT name ON (c:Country) ASSERT c.name is UNIQUE")
+    }
+
+    profileQuery(
+      title = "Drop Named Constraint",
```
Hunterness: Should be `Drop Constraint by name` as for index

cypher/cypher-docs/src/docs/dev/execution-plan-groups/operators.asciidoc
```diff
@@ -169,20 +169,29 @@ include::../ql/query-plan/set-property.asciidoc[]
 
 include::../ql/query-plan/create-unique-constraint.asciidoc[]
 
+[role=deprecated]
```
Lojjs, line 172: Not sure why, but the orange deprecated line for this section is starting in the middle of the section above. That looks weird.

cypher/cypher-prettifier/src/main/scala/org/neo4j/cypher/docgen/tooling/PrettifierParser.scala
```diff
@@ -103,10 +104,14 @@ class PrettifierParser(val keepMyNewlines: Boolean) extends Parser with Base wit
     group(
       keyword("LOAD CSV") |
         keyword("ORDER BY") |
-        keyword("CREATE INDEX ON") |
-        keyword("DROP INDEX ON") |
+        keyword("CREATE INDEX ON") | // Deprecated
+        keyword("CREATE INDEX") |
+        keyword("DROP INDEX ON") | // Deprecated
+        keyword("DROP INDEX") | // These are for the named versions, sadly they will break the query on the ON keyword
```
Lojjs, line 110: I hope we can fix this somehow, it makes it look kind of ugly :(

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/ConstraintTest.scala
```diff
@@ -85,14 +100,26 @@ If any other node with that label is updated or created with a `name` that
...
+
 ###assertion=drop-unique-property-constraint
 //
 
 DROP CONSTRAINT ON (p:Person)
      ASSERT p.name IS UNIQUE
 ###
 
-Drop the unique constraint and index on the label `Person` and property `name`.
+Drop the unique constraint and index on the label `Person` and property `name`. This is deprecated.
```
Lojjs: SInce the refcard should be a quick reference, I think we should remove all the entries that are now deprecated from there. It is enough that they are in the Cypher manual.

Lojjs (changes requested):
Overall it looks good and I also like you improved the existing text a bit! Have some mixed comments though.

cypher/cypher-docs/src/docs/dev/execution-plan-groups/operators.asciidoc
Hunterness: Only sometimes, I have seen the same section both with and without this problem depending on if I clicked a link to get to that section or the one above....

I'm very confused about this

cypher/cypher-prettifier/src/main/scala/org/neo4j/cypher/docgen/tooling/PrettifierParser.scala
Hunterness: I just don't know how....

Lojjs (approved):
I think this is good enough for MR3 :)

mariascharin:
@Hunterness @Lojjs Can I have a look at this today, and then merge?

mariascharin:
@Hunterness @Lojjs 
This is super nice work, thank you!
I have some questions and minor fixes but in the interest of MR3 I will merge this now and follow up when things get a bit more manageable.

PR #442, opened 2019-11-01
Mark which security commands are community/enterprise

Lojjs:
and other small fixes/loose ends to security docs.


mariascharin:
@Lojjs This is super nice! Thank you!

I have one request: Could the subsections under _Graph and Sub-graph Access Control_ and _Security of Administration_ each get the _enterprise-edition_ role please? 

@craigtaverner I think this page is so long now that we ought to chunk it. Do you agree?

Lojjs:
@mariascharin I will fix more enterprise labels for the subsections. I also agree that the page starting to get so long that it would be good to chunk it up to one page per section, not entirely sure how to do it though.

craigtaverner (approved):

PR #444, opened 2019-11-01
Fabric CALL {} and USE fixes

tobias-johansson:
- Better explanation of correlated subqueries
  - Importing variables
  - Aggregation scoping
- Update USE clause doc 
  - Remove invalid example


mariascharin:
Thank you @tobias-johansson !

PR #446, opened 2019-11-04
Section for subquery clauses in clauses overview, and fix CALL {} link

tobias-johansson:


mariascharin:
Thank you @tobias-johansson !

PR #447, opened 2019-11-04
Make use of index and constraint names in Core API examples.

chrisvest:
Also performs a bit of clean-up in those general areas.

Companion PR: https://github.com/neo-technology/neo4j-manual-modeling/pull/923


mariascharin:
Thank you @chrisvest !

PR #459, opened 2019-11-29
4.0 cypher textual cleanup

AlexicaWright:


cypher/cypher-docs/src/docs/dev/general/query-transactions.asciidoc
```diff
@@ -1,9 +1,14 @@
 [[query-transactions]]
 = Transactions
 
+[abstract]
+--
+This section describes transactions with a database.
```
craigtaverner: I think the working here makes more sense for kernel documentation. In this context we are focusing on the interaction between queries and transactions, since users might not know the difference. Perhaps a line like:

>This section describes how Cypher queries work with database transactions.

cypher/cypher-docs/src/docs/dev/general/query-transactions.asciidoc
```diff
@@ -16,14 +21,13 @@ While it is not possible to run a Cypher query outside a transaction, it is poss
 
 Note that the transaction will hold the changes in memory until the whole query, or whole set of queries, has finished executing.
 A query that makes a large number of updates will consequently use large amounts of memory.
-For memory configuration in Neo4j, see the <<operations-manual#memory-configuration, Neo4j Operations Manual>>.
+For memory configuration in Neo4j, see the <<operations-manual#memory-configuration, Neo4j Operations Manual -> Memory configuration>>.
```
craigtaverner, line 24: We originally left out the detail in the link name because it was implied by context. Adding it add redundancy to the sentence. But perhaps you did this because of some other rule you have for inter-document link formatting?

craigtaverner (approved):
Nice set of improvements, both to the older text I did not fix, as well as to my updates. There were one or two things I commented on, but do not want to block this PR.

cypher/cypher-docs/src/docs/dev/general/query-transactions.asciidoc
AlexicaWright: Thank you Craig, this makes perfect sense. I have updated!

cypher/cypher-docs/src/docs/dev/general/query-transactions.asciidoc
AlexicaWright: The latter is the standard way of writing cross links and I feel consistency is more important than the fact that it's a bit redundant given the context.

PR #464, opened 2019-12-03
Fix casing of reltypes.

Lojjs:
Should be null-forward merged (already fixed in 4.0)


craigtaverner (changes requested):
I noticed that Olivia fixed this in 4.0, but also added the query to create the example graph. Perhaps it would be best to cherry-pick her commit back to 3.5, so you get both the CREATE and MATCH parts of this together.

craigtaverner (approved):
My mistake, on further review I see the CREATE part is already there.

PR #468, opened 2019-12-09
Fix bug with broken links

Lojjs:
This was broken in all versions. Should be forward merged all the way to 4.0/4.1.


Lojjs:
Wrong version

PR #477, opened 2019-12-12
Full-text analyzer stop words.

chrisvest:
Mention that the `db.index.fulltext.listAvailableAnalyzers` procedure result includes the list of stop words used by each analyzer.


davidoliverSP2 (approved):
Thanks @chrisvest 👍

PR #478, opened 2019-12-12
Added LOAD CSV functions and updated list comprehension description in the ref card

Hunterness:


cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/ListExpressionsTest.scala
```diff
@@ -120,6 +109,17 @@ RETURN
...
+###assertion=returns-one parameters=value
+MATCH path = (n)-->(m)
+WHERE id(n) = %A% AND id(m) = %B%
+WITH nodes(path) AS list
+RETURN
+
+[x IN list WHERE x.prop <> $value | x.prop]
+###
+
+A list comprehension filters a list and extracts the value of the expression for each element in that list.
```
Lojjs: Since both the above one also are types of list comprehensions, I'm thinking `A list comprehension that filters...` might be better. Now it can be interpreted as if all list comprehension does both filter and extract.

Lojjs (approved):

PR #490, opened 2020-01-10
Describe what ORDER BY does in a WITH clause.

sherfert:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/OrderByTest.scala
```diff
@@ -116,11 +118,26 @@ class OrderByTest extends DocumentingTest {
...
+          r.toList should equal(List(Map("n.age" -> 34, "n.name" -> "A", "n.length" -> 170), Map("n.age" -> 32, "n.name" -> "C", "n.length" -> 185), Map("n.age" -> 36, "n.name" -> "B", "n.length" -> null)))
         })) {
         p("The nodes are returned sorted by the length property, with a node without that property last.")
         resultTable()
       }
     }
+    section("Ordering in a `WITH` clause", "order-with") {
+      p(
+        """When specifying `ORDER BY` in a `WITH` clause, the immediate next clause is allowed to rely on that ordering.
+          |This is useful, e.g. when the next clause contains an aggregation which should receive the rows in a particular order.
```
Mats-SX: I suggest rewording and expanding somewhat:

```
When `ORDER BY` is present on a projection clause (`WITH` or `RETURN`), the immediately following clause will receive records in the specified order.
The order is not guaranteed to be retained after the following clause, unless that also has an `ORDER BY` subclause.
The ordering guarantee can be useful to exploit by operations which depend on the order in which they consume values.
For example, this can be used to control the order of items in the list produced by the `collect()` aggregating function.
The `MERGE` and `SET` clauses also have ordering dependencies which can be controlled this way.
```

Two things that I am not certain on:

1) Is the terminology for aggregations correct above, or should it be 'aggregating functions', 'aggregators' or something else?
2) Is there a part in the manual that discusses the ordering dependencies of `MERGE` and `SET`? If yes, perhaps having a reference here to point there could be useful. I find that this feature is very useful to mitigate the issues that these clauses currently have (and which have been carefully studied in our academic collaborations).

Mats-SX: OK the manual apparently calls it 'aggregating functions' which I think isn't the best, but I suggest to use that term for consistency also here.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/OrderByTest.scala
```diff
@@ -116,11 +118,26 @@ class OrderByTest extends DocumentingTest {
...
+          |""".stripMargin)
+      query(
+        """MATCH (n)
+          |WITH n ORDER BY n.age
+          |RETURN collect(n.name) AS names""".stripMargin, ResultAssertions((r) => {
+          r.toList should equal(List(Map("names" -> List("C", "A", "B"))))
+        })) {
+        p("The list of names built from the `collect` aggregation function contains the names in order of the `age` property.")
```
Mats-SX:
```suggestion
        p("The list of names built from the `collect` aggregating function contains the names in order of the `age` property.")
```

Mats-SX (changes requested):
I had a comment on expanding and rewording the doc a bit.

Also, is the feature already implemented?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/OrderByTest.scala
sherfert: I don't think it makes sense to mention `RETURN` clauses, since there can't be any immediately following clause. I will use the rest of your suggestion.

sherfert:
> Also, is the feature already implemented?

Is has always been, actually. But I will make sure we have tests in Neo4j to make sure it stays that way.

Mats-SX (approved):

Mats-SX:
Aww you didn't co-author me :')

PR #496, opened 2020-01-29
Fixed unresolved db version in Cypher manual

renetapopova:


AlexicaWright (changes requested):
this fixes it, but I also noticed a typo. therefor -> therefore. Well done! :)

PR #497, opened 2020-01-31
Updated the list of additions in 4.0 to also mention existential subqueries

Hunterness:
Also updated section header on existential subqueries to mention the right clause (said WITH but the section was about WHERE)


davidoliverSP2 (approved):
Thanks @Hunterness 👍 

I added some highlighting to the clause names so they would appear in code blocks - I hope this makes sense! please revert if you had intentionally avoided that markup.

Hunterness:
@davidoliverSP2 It looks good, will you merge it?

PR #499, opened 2020-02-03
Backport failing test fix (result order change)

davidoliverSP2:
Do not merge to 4.1 (https://github.com/neo4j/neo4j-documentation/commit/88e5b880a0e0bbd80fcab29f596450910da677ab)

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
@@ -291,13 +291,8 @@ class SecurityUserAndRoleManagementTest extends DocumentingTest with QueryStatis
     }
   })
 
-<<<<<<< HEAD
-  private def assertRolesShown(expected: Seq[String] = List.empty) = ResultAndDbAssertions((p, db) => {
-    val tx = db.beginTransaction(Type.explicit, AnonymousContext.read())
-=======
   private def assertRolesShown(expected: Seq[String] = List.empty, ignore: Seq[String] = List.empty) = ResultAndDbAssertions((p, db) => {
     val tx = db.beginTransaction(Type.EXPLICIT, AnonymousContext.read())
->>>>>>> 88e5b880a0... Fixed failing tests due to change in result order
     try {
       val result = p.columnAs[String]("role").toList.filter(!ignore.contains(_))
       result.toSet should equal(expected.toSet)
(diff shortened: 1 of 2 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
```diff
@@ -282,23 +282,20 @@ class SecurityUserAndRoleManagementTest extends DocumentingTest with QueryStatis
...
+      val props = nodes.map(n => n.getProperty("name"))
+      val result = p.columnAs[String](column).toList
+      result.toSet should equal(props.toSet)
     } finally {
       tx.close()
     }
   })
 
-  private def assertRolesShown(expected: Seq[String] = List.empty) = ResultAndDbAssertions((p, db) => {
-    val tx = db.beginTransaction(Type.explicit, AnonymousContext.read())
```
davidoliverSP2, line 298: @craigtaverner I had to fix a merge conflict here (see the 2nd commit) - just wanted to check with you that this isn't going to cause problems

craigtaverner: My commit did not change the line for beginTransaction, so I assume whatever was the version for 4.0 should be the right one. For 4.1 I re-ran all tests before pushing. If all tests pass here, we should be fine. This test method is only used in one test, so it might be worth reviewing the output of that as well, in case it has changed. One suspicion I have is the addition of the PUBLIC role, which is a 4.1 only change, will make the output table look cluttered. This is completely independent of this commit, but worth considering when reviewing the 4.1 docs.

craigtaverner (approved):
I think this looks good. The one line different is, I presume, something that is needed for 4.0.

craigtaverner (changes requested):
I changed my mind after looking at the build failure:

```
SecurityUserAndRoleManagementTest.scala:295: value EXPLICIT is not a member of object org.neo4j.kernel.api.KernelTransaction.Type
```

So the merge conflict was not correctly changed. I think you need to revert that one line back to the 4.0 version. I strongly recommend also compiling it locally and running all tests before merging this PR.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
craigtaverner: See my final comment below. The beginTransaction line is wrong.

davidoliverSP2:
thanks @craigtaverner 

I reset the `beginTransaction` line to how it was in 4.0, and compiled it locally without any problems. The tests on here seem to be happy too.

craigtaverner (approved):
Super!

PR #503, opened 2020-02-06
Updated the list of deprecations and removals in 4.0 to be correct for variable binding of variable length relationships

Hunterness:
It was decided to keep it deprecated and not remove it for now.


davidoliverSP2 (approved):
Thanks @Hunterness 👍 

I'll merge

PR #516, opened 2020-02-27
Added info about revoking compound privileges also revokes sub-privileges

Hunterness:
Should not be forwarded to 4.1 since it works differently there.

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityPrivilegesTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityPrivilegesTest.scala
@@ -63,7 +63,7 @@ class SecurityPrivilegesTest extends DocumentingTest with QueryStatisticsTestSup
             |* action: which type of privilege this is: access, traverse, read, write, token, schema or admin
             |* resource: what type of scope this privilege applies to: the entire dbms, a database, a graph or sub-graph access
             |* graph: the specific database or graph this privilege applies to
-            |* segement: for sub-graph access control, this describes the scope in terms of labels or relationship types
+            |* segment: for sub-graph access control, this describes the scope in terms of labels or relationship types
             |* role: the role the privilege is granted to
             |""".stripMargin)
         resultTable()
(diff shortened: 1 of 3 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityPrivilegesTest.scala
```diff
@@ -224,20 +232,34 @@ class SecurityPrivilegesTest extends DocumentingTest with QueryStatisticsTestSup
...
       query("REVOKE TRAVERSE ON GRAPH neo4j NODES Payments FROM regularUsers", ResultAssertions((r) => {
-        // TODO: revoke a privilege that exists
-        assertStats(r, systemUpdates = 0)
+        assertStats(r, systemUpdates = 2)
       })) {
         statsOnlyResultTable()
       }
+
+      p(
+        """Some privileges are compound privileges and contains sub-privileges, for example <<administration-security-administration-database-indexes, `INDEX MANAGEMENT`>> which covers `CREATE INDEX` and `DROP INDEX`.
```
Hunterness, line 249: Should we mention that you can't revoke a sub-privilege when a compound version of the compound privilege exists?

craigtaverner: The tricky part is that in theory all compound privileges should have been split to leaves on creation. This is true for all that were created using Cypher, but not true for some of those created in the system graph initialization. So this problem is specific to system graph initialization privileges only, and perhaps needs to be described in the operations manual, not here where we focus on Cypher commands only (and there are no bugs).

craigtaverner (approved):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityPrivilegesTest.scala
Hunterness: `This is true for all that were created using Cypher` Except for ROLE MANAGEMENT :P

davidoliverSP2 (approved):
thanks @Hunterness 👍

PR #522, opened 2020-03-09
Fix tests.

sherfert:
Fixes a failing test.


alexaverbuch (approved):
looks good, as long as we don't forget to clean up this `runtime=slotted` ugliness when we come to documenting `Anti` and `MultiNodeIndexSeek` (as discussed)

PR #525, opened 2020-03-17
Document `Anti` and `MultiNodeIndexSeek`

sherfert:


davidoliverSP2 (approved):
Looks good, thanks @sherfert 👍 

I think this page would benefit from a sub-menu, so that users can see a full list of the operators. That doesn't have to be done in this PR though.

If this isn't dependant on any other PR, I'm happy to merge

PR #527, opened 2020-03-23
Add TransactionEventListener[] example tags

MishaDemianenko:


davidoliverSP2:
Merged forward from #528

PR #528, opened 2020-03-23
Add TransactionEventListener[] example tags

MishaDemianenko:


davidoliverSP2 (approved):
Thanks @MishaDemianenko 👍

PR #551, opened 2020-04-06
Document new user management privilege commands.

Lojjs:
Documentation of the new  `GRANT SET PASSWORDS ON DBMS TO role` and `GRANT SET STATUS ON DBMS TO role` user management commands.


SaschaPeukert (changes requested):
Apart from the "what is the best plural of 'status' question, I would recommend adding examples that show how granting those privileges actually affect the `ALTER USER` commands. Also: we explain `ALTER USER` in other parts of the docs but should add a cross-reference ("You need to be granted XYZ in order to execute this") to the new privileges there now

SaschaPeukert (approved):
Since the rest of my points are extracted into another card, I am happy with this now

davidoliverSP2 (approved):
Thanks @Lojjs 👍

PR #552, opened 2020-04-07
fixed typo and updated column mention

AlexicaWright:


renetapopova (approved):
Looks good!  💃

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
```diff
@@ -160,21 +160,22 @@ class SecurityUserAndRoleManagementTest extends DocumentingTest with QueryStatis
...
           resultTable()
         }
         p("There are multiple versions of this command, the default being `SHOW ALL ROLES`. " +
           "To only show roles that are assigned to users, the command is `SHOW POPULATED ROLES`. " +
           "To see which users are assigned to roles `WITH USERS` can be appended to the commands. " +
-          "This will give one result row for each user, so if a role is assigned to two users then it will show up twice in the result.")
+          "This will give one result row for each user, so if a role is assigned to two users then it will show up twice in the result. ")
         query("SHOW POPULATED ROLES WITH USERS", assertRolesShown(Seq("admin"), Seq("PUBLIC"))) {
-          p("The table of results contains two columns, the first is the role name, and the other a flag indicating whether the role is a built-in role or a custom role.")
+        p("The table of results will show information about the role, i.e. its name, whether or not it is built-in and what database it belongs to. ")
```
davidoliverSP2: I think it might be a good idea not to reference `isBuiltIn` at all, due to the Cypher team wanting to remove this column.

Perhaps something like:
> The table of results will show information about the role and what database it belongs to.

davidoliverSP2 (changes requested):
Thanks @AlexicaWright 👍 

I think we should avoid referencing `isBuiltIn` directly, as it may be removed in the near future.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
AlexicaWright: Agree!

PR #554, opened 2020-04-08
Removed ELEMENTS[S] from WRITE command syntax

renetapopova:


davidoliverSP2 (approved):
Thanks @renetapopova - looks good.

I think we should just double check this with someone in the Cypher team - perhaps you could ask Louise? (maybe you already have?)

renetapopova:
> Thanks @renetapopova - looks good.
> 
> I think we should just double check this with someone in the Cypher team - perhaps you could ask Louise? (maybe you already have?)

Yes, I've done that.

AlexicaWright (commented):
I agree with @davidoliverSP2 , it looks good to me, but maybe Louise or someone else in her team can review, just to be on the safe side?

Lojjs (approved):
Looks good to me. Forward merge it to 4.1 for now and then if we (fingers crossed) manage to implement fine grained writes in 4.1 we can expand this docs section at that point

PR #558, opened 2020-04-21
Remove unused dependency

klaren:


klaren:
@davidoliverSP2 The `EmbeddedNeo4jWithSLF4JLogging.java` is removed since 4.0 so maybe we should remove it there as well. We already published the maven module for 4.0 so we need to keep it there for 4.0.x, but I'm removing the maven module for 4.1+.

https://github.com/neo4j/neo4j-documentation/blob/3.5/embedded-examples/src/main/java/org/neo4j/examples/EmbeddedNeo4jWithSLF4JLogging.java

klaren:
For reference https://github.com/neo-technology/neo4j/pull/5505

PR #561, opened 2020-04-24
Add example of combination of variable length and multiple relationship types.

Lojjs:
Also fixed a small typo that I happened to see.


davidoliverSP2 (approved):
Thanks @Lojjs - this looks good 👍

PR #563, opened 2020-04-28
Check plan description arguments when validating query details 

LinneaAndersson:
This PR fixes the failures in the build, see https://live.neo4j-build.io/buildConfiguration/Documentation_Neo4jDocumentationTests_Neo4j41DocsTest/7556982?

### Description of change
The plan description table was changed to truncate the column that renders the query details to the first 100 characters. Instead of validating the rendering of the table, we should validate the arguments in the InternalPlanDescription.


renetapopova (approved):
Thanks a lot, @LinneaAndersson!

PR #564, opened 2020-05-02
Sub-menu in the Operators page

renetapopova:
Replaces #562

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/execution-plan-groups/operators.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/execution-plan-groups/operators.asciidoc
@@ -7,6 +7,97 @@ All operators are listed here, grouped by the similarity of their characteristic
 Certain operators are only used by a subset of the <<cypher-runtime,runtimes>> that Cypher can choose from.
 If that is the case, the example queries will be prefixed with an option to choose one of these runtimes.
 
+* <<query-plan-all-nodes-scan, AllNodesScan>>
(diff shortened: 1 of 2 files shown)


AlexicaWright (commented):
Thank you @renetapopova , lots of work! As per the discussion on the Trello card, I am on the fence whether this is an improvement or not. There are so many operators that one doesn't get a good overview even with a sub-menu. However, one do get a sense of how many there are and I guess that counts. Bottom line, I say let's merge this!

renetapopova:
@davidoliverSP2 please review it

cypher/cypher-docs/src/docs/dev/execution-plan-groups/operators.asciidoc
```diff
@@ -1,12 +1,104 @@
 [[execution-plans-operators]]
-= Operators
+= Execution plan operators in details
```
davidoliverSP2: I think this should be 

> Execution plan operators in detail

davidoliverSP2 (changes requested):
Hi @renetapopova - this looks good, but there is something odd going on with the index file in this section - if you build this and look at Chapter 7. Execution plans, there is a lot of content building, and the toc doesn't include this page.

davidoliverSP2 (approved):
Thanks @renetapopova - this is excellent 👍 

There is still an oddity with the back and forward links, so I created https://github.com/neo-technology/neo4j-manual-modeling/pull/1122 to update the content map.

PR #567, opened 2020-05-11
add performance info about merge

AlexicaWright:


cypher/cypher-docs/src/docs/dev/ql/merge/index.asciidoc
```diff
@@ -36,6 +36,7 @@ Either the pattern already exists, or it needs to be created.
 
 `MERGE` either matches existing nodes and binds them, or it creates new data and binds that.
 It's like a combination of `MATCH` and `CREATE` that additionally allows you to specify what happens if the data was matched or created.
+For performance reasons, creating a schema index on the label or property is highly recommended when using `MERGE`.
 
 For example, you can specify that the graph must contain a node for a user with a certain name.
```
davidoliverSP2: The positioning of this looks a little odd to me - the example could be seen as an example of creating a schema index on the label/property, which it's not. 

I think this would probably be a good case for an admonition, perhaps placed after the example paragraph with a link to the Indexes for search performance section:

> For example, you can specify that the graph must contain a node for a user with a certain name. If there isn’t a node with the correct name, a new node will be created and its name property set.
>
>[NOTE]
>For performance reasons, creating a schema index on the label or property is recommended when using `MERGE`. For more information, see Indexes for search performance (administration-indexes-search-performance).

davidoliverSP2 (changes requested):
Thanks @AlexicaWright 

Any idea why this content is duplicated in both the scala test, and the asciidoc folder?

I think the note about performance is a useful tip and would benefit from highlighting in an admonition, with a link to the performance page in the Administration section.

davidoliverSP2 (approved):

PR #569, opened 2020-05-12
Added a note that clarifies WITH

renetapopova:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/WithTest.scala
```diff
@@ -41,6 +41,9 @@ class WithTest extends DocumentingTest {
                 (b)-[:BLOCKS]->(d)
       """.stripMargin)
     synopsis("The `WITH` clause allows query parts to be chained together, piping the results from one to be used as starting points or criteria in the next.")
+    note {
+      p("Bear in mind that `WITH` affects only variables in scope. Variables that are not included in the `WITH` clause will not be carried over to the rest of the query.")
```
davidoliverSP2: I would suggest a slight rewording:

> It is important to note that the `WITH` clause affects variables in scope. Any variables not included in the `WITH` clause are not carried over to the rest of the query.

davidoliverSP2 (approved):
Thanks @renetapopova - this is a useful addition. I made a small recommendation to reword it, based on the information in the card, but I would also recommend asking a member of the Cypher team to review this for accuracy.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/WithTest.scala
renetapopova: So, you suggest the note to be as in the card. I reworded it to "It is important to note that `WITH` affects variables in scope. Any variables not included in the `WITH` clause are not carried over to the rest of the query."

davidoliverSP2: 👍  yep, this looks good.

sherfert (approved):

renetapopova:
Thank you @sherfert !

PR #577, opened 2020-05-19
Removed the traversal doc test

renetapopova:


davidoliverSP2 (approved):
This builds cleanly as far as I can tell! thanks @renetapopova

PR #581, opened 2020-05-25
Update glossary.asciidoc

sherfert:
This is not up to date at all.

Is there a checklist item missing on card to update this for each release?


AlexicaWright (approved):
Looks good, however, a suggestion would be to add the "enterprise edition" tag. Thank you for updating this @sherfert , looks like it was long overdue!

PR #586, opened 2020-05-27
Improved the ACCESS privilege definition

renetapopova:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityAdministrationTest.scala
```diff
@@ -61,8 +61,8 @@ class SecurityAdministrationTest extends DocumentingTest with QueryStatisticsTes
       // image source: https://docs.google.com/drawings/d/1Zc5eawW58r4CI8e4FeYnA8cwUyIJX-7B4Eecj-qkd_8/edit
       section("The database `ACCESS` privilege", "administration-security-administration-database-access", "enterprise-edition") {
         p(
-          """The `ACCESS` privilege can be used to enable the ability to access a database.
-            |If this is not granted to users, they will not even be able to start transactions on the relevant database.""".stripMargin)
+          """The `ACCESS` privilege enables users to read data and perform updates in a database.
```
craigtaverner: The access privilege does not provide any read or write privileges, so this sentence is false. It allows connecting to the database only, and any reads or writes will not work (reads will return no results and writes will cause an error).

craigtaverner (changes requested):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityAdministrationTest.scala
renetapopova: Thanks @craigtaverner, I see what you mean. I've misinterpreted the sentence "If this is not granted to users, they will not be able to read any data, or perform any updates to the database."

craigtaverner (approved):
Much better. I wonder if it would not be a good idea to also mention that it is insufficient to allow reading and writing the database? I think that can be deduced from this content, but sometimes it helps to be explicit.

davidoliverSP2 (approved):
👍  thanks @renetapopova

PR #589, opened 2020-05-27
Document new write privileges

Lojjs:
Added to:
- Security write section
- Glossary
- Additions
- Refcard

Needs to go in together with https://github.com/neo-technology/neo4j-manual-modeling/pull/1153 which adds the updated privilege image and https://github.com/neo-technology/neo4j-manual-modeling/pull/1154 which adds the new sections to the Cypher refcard.


cypher/cypher-docs/src/docs/dev/ql/administration/security/index.asciidoc
```diff
@@ -30,7 +30,14 @@ This section explains how to use Cypher to manage Neo4j role-based access contro
...
 ** <<administration-security-reads-match, The `MATCH` privilege>>
 * <<administration-security-writes, Write privileges>>
+** <<administration-security-writes-create, The `CREATE` privilege>>
+** <<administration-security-writes-delete, The `DELETE` privilege>>
+** <<administration-security-writes-set-label, The `SET LABEL` privilege>>
+** <<administration-security-writes-remove-label, The `REMOVE LABEL` privilege>>
+** <<administration-security-writes-set-property, The `SET PROPERTY` privilege>>
+** <<administration-security-writes-merge, The `MERGE` privilege>>
 ** <<administration-security-writes-write, The `WRITE` privilege>>
+** <<administration-security-writes-all, `ALL GRAPH PRIVILEGES`>>
```
davidoliverSP2: While this looks odd, should this section be titled 

The `ALL GRAPH PRIVILEGES` privilege

Just for consistency with the other headers?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+        |* `DELETE` - allows deleting nodes and relationships
+        |* `SET LABEL labels` - allows setting the specified node labels using the `SET` clause.
+        |* `REMOVE LABEL labels` - allows removing the specified node labels using the `REMOVE` clause.
+        |* `SET PROPERTY +{props}+` - allows setting properties on nodes and relationships.
+        |
+        |There are also compound privileges which combine the above specific privileges:
+        |
+        |* `MERGE` - allows match, create and set property to permit the `MERGE` command.
+        |* `WRITE` - allows all write operations on an entire graph.
+        |* `ALL GRAPH PRIVILEGES` - allows all read and write operation on an entire graph.
```
davidoliverSP2: I think in each of these bullets, "allows" can be replaced with "enables".
Also, some bullets are missing full stops.

If it's helpful, I don't mind pushing editorial changes directly to the PR - ping me if you'd prefer that.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+      p("include::deny-create-syntax.asciidoc[]")
+
+      p("For example, denying the ability to create nodes with the label `foo` on all graphs to the role `regularUsers` would be achieved using:")
+      query("DENY CREATE ON GRAPH * NODES foo TO regularUsers", ResultAssertions(r => {
+        assertStats(r, systemUpdates = 1)
+      })) {
+        statsOnlyResultTable()
+      }
+      note {
+        p("If the users attempts to create nodes with a label that does not already exist on the database the user must also possess the " +
```
davidoliverSP2:
```suggestion
        p("If the user attempts to create nodes with a label that does not already exist on the database, then the user must also possess the " +
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+
+      p("For example, denying the ability to delete relationships with the relationship type `bar` on all graphs to the role `regularUsers` would be achieved using:")
+      query("DENY DELETE ON GRAPH * RELATIONSHIPS bar TO regularUsers", ResultAssertions(r => {
+        assertStats(r, systemUpdates = 1)
+      })) {
+        statsOnlyResultTable()
+      }
+      note {
+        p(
+          """Users with `DELETE` privilege but restricted `TRAVERSE` privileges will not be able to do `DETACH DELETE` in all cases.
```
davidoliverSP2:
```suggestion
          """Users with `DELETE` privilege, but restricted `TRAVERSE` privileges, will not be able to do `DETACH DELETE` in all cases.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+        |There are also compound privileges which combine the above specific privileges:
+        |
+        |* `MERGE` - allows match, create and set property to permit the `MERGE` command.
+        |* `WRITE` - allows all write operations on an entire graph.
+        |* `ALL GRAPH PRIVILEGES` - allows all read and write operation on an entire graph.
         |""".stripMargin)
 
+    section("The `CREATE` privilege", "administration-security-writes-create", "enterprise-edition") {
+      p(
+        """The `CREATE` privilege enables a user to create new node and relationship elements in a graph. See <<query-create, CREATE>>""".stripMargin)
```
davidoliverSP2:
```suggestion
        """The `CREATE` privilege enables a user to create new node and relationship elements in a graph. See <<query-create, CREATE>>.""".stripMargin)
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+
+      p(
+        """For example, granting `MERGE` on all elements of the graph `neo4j` to the role `regularUsers` would be achieved using:""".stripMargin)
+      query("GRANT MERGE {*} ON GRAPH neo4j ELEMENTS * TO regularUsers", ResultAssertions(r => {
+        assertStats(r, systemUpdates = 2)
+      })) {
+        statsOnlyResultTable()
+      }
+
+      p("It is not possible to deny the `MERGE` privilege. If it is desirable to prevent a users from creating elements and setting properties use " +
```
davidoliverSP2:
```suggestion
      p("It is not possible to deny the `MERGE` privilege. If it is desirable to prevent a users from creating elements and setting properties, use " +
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -14,23 +14,208 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+          "<<administration-security-administration-database-tokens, CREATE NEW LABEL>> privilege. The same applies to new relationships and properties - the " +
+          "<<administration-security-administration-database-tokens, CREATE NEW RELATIONSHIP TYPE>> or " +
+          "<<administration-security-administration-database-tokens, CREATE NEW PROPERTY NAME>> privileges are required.")
+      }
+    }
+
     section("The `WRITE` privilege", "administration-security-writes-write", "enterprise-edition") {
       p(
-        """The `WRITE` privilege enables you to write on a graph.""".stripMargin)
+        """The `WRITE` privilege allows the user to execute any write command on a graph.""".stripMargin)
```
davidoliverSP2, line 224:
```suggestion
        """The `WRITE` privilege enables the user to execute any write command on a graph.""".stripMargin)
```

davidoliverSP2: I see this was previously "enables". 

In the context of privileges, the difference between enabling and allowing is thin. I think we should default to consistency.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -41,7 +226,8 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
         statsOnlyResultTable()
       }
       note {
-        p("Unlike with `GRANT READ` it is not possible to restrict `WRITE` privileges to specific ELEMENTS, NODES or RELATIONSHIPS.")
+        p("Unlike the more specific write commands, it is not possible to restrict `WRITE` privileges to specific ELEMENTS, NODES or RELATIONSHIPS. If it is " +
+        "desirable to prevent a user from writing to a subset of database objects, a `GRANT WRITE` can be combined with more specific DENYs targeting these elements.")
```
davidoliverSP2:
```suggestion
        "desirable to prevent a user from writing to a subset of database objects, a `GRANT WRITE` can be combined with more specific `DENY` commands to target these elements.")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -59,6 +245,34 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
             | See <<operations-manual#detach-delete-restricted-user, Operations Manual -> Fine-grained access control>> for more info.""".stripMargin)
       }
     }
+
+    section("`ALL GRAPH PRIVILEGES`", "administration-security-writes-all", "enterprise-edition") {
+      p(
+        """The `ALL GRAPH PRIVILEGES` privilege allows the user to execute any command on a graph.""".stripMargin)
```
davidoliverSP2, line 259:
```suggestion
        """The `ALL GRAPH PRIVILEGES` privilege enables the user to execute any command on a graph.""".stripMargin)
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
```diff
@@ -59,6 +245,34 @@ class SecurityWritePrivilegesTest extends DocumentingTest with QueryStatisticsTe
...
+      p(
+        """For example, granting all graph privileges on the graph `neo4j` to the role `regularUsers` would be achieved using:""".stripMargin)
+      query("GRANT ALL GRAPH PRIVILEGES ON GRAPH neo4j TO regularUsers", ResultAssertions(r => {
+        assertStats(r, systemUpdates = 1)
+      })) {
+        statsOnlyResultTable()
+      }
+      note {
+        p("Unlike the more specific read and write commands, it is not possible to restrict `ALL GRAPH PRIVILEGES` privileges to specific ELEMENTS, NODES or RELATIONSHIPS. If it is " +
+          "desirable to prevent a user from reading or writing to a subset of database objects, a `GRANT ALL GRAPH PRIVILEGES` can be combined with more specific DENYs targeting these elements.")
```
davidoliverSP2:
```suggestion
          "desirable to prevent a user from reading or writing to a subset of database objects, a `GRANT ALL GRAPH PRIVILEGES` can be combined with more specific `DENY` commands to target these elements.")
```

davidoliverSP2 (changes requested):
Thanks @Lojjs - this looks great. Most of my comments are editorial suggestions. 

Throughout the manuals we're a little inconsistent with usage of allows vs enables. In most cases it usually should be "enables", but this section is specifically referencing permissions, so it falls into a bit of a grey area. I would recommend we stay consistent where we can though.

I've not yet build the Refcard - will address in further comments when I get that to build locally.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
craigtaverner: When thinking about the difference between 'allows' and 'enables' I cannot help but think that if something is 'not allowed' that is an error, while if something is 'not enabled' it just means nothing will happen. Now when we compare READ and WRITE privileges, the READ privileges will simply control what you see (with no error) and so better match the word _enables_, while WRITE privileges can result in actual errors (you get an exception if you try write something you are not allowed to), which certainly better matches the word _allows_. Would it make sense to make use of _BOTH_ words then, but with consistent association with these two very different cases? This will also help maintain clarity for the user that READ and WRITE privileges behave in such a very different way.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
craigtaverner: I think we have an opportunity to highlight a key difference between READ and WRITE privileges. See my comment above. In this particular case we should use _allows_ because we would result in an error without it.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
davidoliverSP2: thanks @craigtaverner for the clarity - I'm in agreement, let's use both where appropriate.

Additionally, when we finally get to building a documentation styleguide, these descriptions of when to use "enables" and "allows" are very clear! Many thanks!

davidoliverSP2:
[GRAPH (WRITE) Privileges](https://trello.com/c/tDrxBGsE/80-graph-write-privileges)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
Lojjs: I like that distinction as well @craigtaverner. We should probably go through the docs later though and see that it is consistent everywhere since we have not written them with this in mind.

cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-create-syntax.asciidoc
```diff
@@ -0,0 +1,9 @@
+.Command syntax
+[source, cypher]
+-----
+DENY CREATE ON GRAPH[S] { * | graph-name[,...] }
+    [ ELEMENT[S] { * | label-or-rel-type-name[,...] }
```
craigtaverner: The formatting here is a little hard to read. We did it in a more open way in the fine-grained read examples.

craigtaverner (changes requested):
The 4.0 documents have security sub-section on their own page, but in this PR they are all bundled into one very long page. Is this deliberate, or an oversight?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityWritePrivilegesTest.scala
craigtaverner: I've made a commit (will push soon) that defines _enable_ and _allow_ in the previous section (fine-grained) and then uses _enable_ in the reads section and _allow_ in the writes section.

craigtaverner (approved):

davidoliverSP2 (approved):

PR #607, opened 2020-06-09
4.1 fabric administration updates

davidoliverSP2:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/DatabasesTest.scala
```diff
@@ -29,9 +29,13 @@ class DatabasesTest extends DocumentingTest with QueryStatisticsTestSupport {
...
-          |including the associated security model, is maintained in a special database called the `system` database.
-          |All multi-database administrative commands need to be executing against the `system` database.""".stripMargin)
+        """Neo4j supports the management of multiple databases within the same DBMS.
+          |The metadata for these databases, including the associated security model, is maintained in a special database called the `system` database.
+          |Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system.
+          |This includes all commands related to managing multiple databases as well as all commands for defining the security model: users, roles and privileges.
+          |The administrative commands that are specific to the schema of an individual database are still executed against that specific database.
+          |These include index and constraint management commands.
+          |When connected to the DBMS over bolt, administrative commands are automatically routed to the `system` database.""".stripMargin)
```
Hunterness: Since we have just mentioned administrative commands that shouldn't be run against `system` maybe we should rephrase the sentence a bit here.

Might add the comment on the `Most administrative commands must be executed against the `system` database ...` part instead => 

```
Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system. These are automatically routed to the `system` database when connected to the DBMS over bolt.
```

Hunterness: i know we didn't think of this when giving our comments in the google doc but I thought of it now :P

Hunterness: I think this section is more an introduction to ALL administration commands, Database, Index, Constraint and Security and not just the Database where you have moved it (might be good to confirm this move with (current) Team Cypher Implementation if you think it fits better here)

cypher/cypher-docs/src/docs/dev/administration.asciidoc
```diff
@@ -42,14 +42,6 @@ This section explains how to use Cypher to administer Neo4j databases, such as c
...
 
 
-Neo4j supports the management of manage multiple databases within the same DBMS.
-The metadata for these databases, including the associated security model, is maintained in a special database called the `system` database.
-Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system.
-This includes all commands related to managing multiple databases as well as all commands for defining the security model: users, roles and privileges.
-The administrative commands that are specific to the schema of an individual database are still executed against that specific database.
-These include index and constraint management commands.
-
-
```
Hunterness, line 52: See my other comment about the move of this introduction

Hunterness (changes requested):
Wants at least some discussion on the moved introduction before approving :)

davidoliverSP2:
> Wants at least some discussion on the moved introduction before approving :)

Thanks @Hunterness - I returned the introduction text to the top of the section. This is going to be reviewed in an upcoming PR to re-architect this section of the docs, but I agree it was moved away in error.

In the _Databases_ intro, I returned it to more or less the content that was there before, editing slightly to mirror the wording in the main introduction. Again, this is likely to be revisited as part of a larger restructure.

Hunterness (approved):
I can see that the section moved higher up in the file but I guess that is fine (and might be changed in a restructure later anyway so don't particularly care now :P)

PR #620, opened 2020-07-06
Add `SHOW ROLES name1, name2 PRIVILEGES` and `SHOW USERS name1, name2 PRIVILEGES`

Hunterness:


cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
```diff
@@ -16,9 +16,9 @@ SHOW USERS
 
 | [source, cypher]
 ----
-SHOW USER [name] PRIVILEGES
+SHOW USER[S] [name1[, name2]] PRIVILEGES
 ----
-| List the privileges granted to a user.
+| List the privileges granted to the specified users, or the current user if omitted.
```
hedengran: Do you think that it would be clearer by writing something like "...or the current user, if no user is specified"? To me it's not clear what can be omitted right now.

Hunterness: The `[...]` defines it as optional, and here its for the whole `[name1[, name2]]` but I agree that it can be confusing, will update :)

hedengran (approved):

davidoliverSP2 (approved):
LGTM, I'll merge 👍 

Thanks @Hunterness

PR #621, opened 2020-07-07
Add YIELD and WHERE to SHOW commands in 4.1 additions section 

mnd999:
Add WHERE and YIELD changes to SHOW commands to deprecations and additions page


Lojjs (approved):
I haven't actually build it but it looks alright to me

PR #624, opened 2020-07-16
Prevent test assertion overwriting the generated request/response for …

gjmwoods:
Prevent test assertion overwriting the generated request/response example for rolling back an http-api transaction.


AlexicaWright (approved):
Builds fine and resolves the issue. Thank you @gjmwoods

PR #626, opened 2020-07-16
Add default graph keyword to syntax, examples and refcard

hedengran:


Hunterness:
I'm a bit confused as to why I have commits here?

hedengran:
That is an excellent question @Hunterness. When I branched for this PR, I accidentally branched from your PR which I reviewed. I assume that it doesn't matter?

Hunterness:
@hedengran probably not but I think it might be better to rebase on (or reset on top of) latest 4.2

It will also help Sascha know what things he need to review and which are old (but I guess all those with your face is for him so  :P)

hedengran:
Wow, that was not better. Will sort this out.

SaschaPeukert (approved):

PR #631, opened 2020-07-20
Add TransactionEventListener tags to example code

MishaDemianenko:
Add TransactionEventListener tags for example to be included into documentation


MishaDemianenko:
Hi @renetapopova i think this would be a way to include an example in a doc. Unfortunately can't check how it looks like in a final doc since can't build that locally

renetapopova (approved):
Thanks @MishaDemianenko! Looks goods and works good 👍

renetapopova:
@MishaDemianenko, could you please cherry-pick it in 4.1.

PR #632, opened 2020-07-22
update round function with new parameters

hedengran:


SaschaPeukert (commented):
I like what you did but I got a question: From what I read here, I would assume that you can not call round with a number and a mode alone. Is that so?

hedengran:
> I like what you did but I got a question: From what I read here, I would assume that you can not call round with a number and a mode alone. Is that so?

Yes, that is correct. Calling round with only a rounding mode is not supported.

SaschaPeukert:
> > I like what you did but I got a question: From what I read here, I would assume that you can not call round with a number and a mode alone. Is that so?
> 
> Yes, that is correct. Calling round with only a rounding mode is not supported.

Alright. That's a possible problem. But not the fault of this PR. I am gonna raise a card to fix this

SaschaPeukert (approved):

PR #637, opened 2020-07-29
4.0 Limitations of security

craigtaverner:
Added section to describe how indexes take into account the security rules. The motivation for this work is that full-text indexes cannot completely take into account blacklisted properties because we cannot introspect the lucene query and as a result we will return empty results for any full-text index that _might_ contain blacklisted results that are not possible to detect.


SaschaPeukert:
I've picked this card up and included some other limitations of role-based access control and created a whole page collecting these. This is now way broader than the original PR but solves 3 different cards. To get this content out the door, we can start having it statically, but may revisit it to make it dynamically generated.

SaschaPeukert:
Fully build docs [4.0-preview.zip](https://github.com/neo4j/neo4j-documentation/files/5181501/4.0-preview.zip)

cypher/cypher-docs/src/docs/dev/administration.asciidoc
```diff
@@ -38,9 +38,10 @@ This section explains how to use Cypher to administer Neo4j databases, such as c
 ** <<administration-security-users-and-roles, User and role management>>
 ** <<administration-security-subgraph, Graph and subgraph access control>>
 ** <<administration-security-administration, Security of administration>>
+** <<administration-security-limitations, Known limitations of security>>
 
 
-Neo4j supports the management of manage multiple databases within the same DBMS.
+Neo4j supports the management of multiple databases within the same DBMS.
```
Lojjs, line 44: Good catch

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
+=== How does role-based access control impact label visibility
+
+==== Traversing the graph with multi-labeled nodes
```
Lojjs: I don't really like that you have 2 headlines after each other here. With the `5.5.5.2. Security and labels`  headline, that means 3 headlines in a row. I think there are too many levels of sections even without this, plus that it looks a bit like there is missing text between the headlines. 

Perhaps you could merge "Security and labels" and "How does role-based access control impact label visibility" into one single headline and then change the other headlines of the section to have 3 instead of 4 ===?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityKnownLimitationsTest.scala
```diff
@@ -0,0 +1,33 @@
...
+    database("neo4j")
+    initQueries(
+      "CREATE INDEX foo FOR (n:Person) ON (n.name)"
+    )
+    //TODO: Make this page dynamic
+    synopsis("This section explains known limitations and implications of Neo4js role-based access control security.")
+
+    section("Security and indexes", "administration-security-limitations-indexes") {
+
+      section("How do privileges impact index results") {
```
Lojjs: Same as for the Security and Labels section, I don't like that it starts with 2 different headlines, both "5.5.5.1. Security and indexes" and "How do privileges impact index results". I think it would be better if you could merge them into one title and move the other headlines of the section one step up.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
+The rules of a security model can have some impact on some of the database operations.
```
AlexicaWright:
```suggestion
The rules of a security model may impact some of the database operations.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+This comes down to necessary additional security checks that incur additional data accesses.
+Especially when it comes to count store operations, as they are usually very fast lookups, this difference might be noticeable.
```
AlexicaWright:
```suggestion
Especially in regards to count store operations, as they are usually very fast lookups, the difference might be noticeable.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityKnownLimitationsTest.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.neo4j.cypher.docgen.tooling._
+
+class SecurityKnownLimitationsTest extends DocumentingTest with QueryStatisticsTestSupport {
+  override def outputPath = "target/docs/dev/ql/administration/security/"
+
+  override def doc: Document = new DocBuilder {
+    doc("Known limitations of security", "administration-security-limitations")
+    database("neo4j")
+    initQueries(
+      "CREATE INDEX foo FOR (n:Person) ON (n.name)"
```
Lojjs: I guess this is only need if we decide to make the page dynamic?

cypher/cypher-docs/src/docs/dev/ql/administration/security/index.asciidoc
```diff
@@ -40,6 +40,7 @@ This section explains how to use Cypher to manage Neo4j role-based access contro
 ** <<administration-security-administration-dbms-privileges, DBMS administration>>
 *** <<administration-security-administration-dbms-custom, Creating custom roles with DBMS privileges>>
 *** <<administration-security-administration-dbms-privileges-role-management, The dbms `ROLE MANAGEMENT` privileges>>
+* <<administration-security-limitations, Known limitations of security>>
```
Lojjs, line 43: The other sections in this menu has links to the subsections as well. Do we want to do the same and add links to the three parts of your new section?

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
+As described in <<administration-indexes-search-performance, Indexes for search performance>>, Neo4j {neo4j-version} supports the creation and use of indexes to improve the performance of Cypher queries.
+The Neo4j security model will impact the results of queries that use the index.
+When using normal Neo4j indexes, a Cypher query will always return the same results it would have if no index existed.
```
Lojjs: I think this could be better formulated
1. What is meant by `normal` Neo4j indexes? I guess it refers to non full-text indexes?
2. The Neo4j security model will impact the results of queries regardless of if they use the index or not. Queries with and without indexes should return the same result regardless of if there are any security restrictions or not. I think that these two sentences can give another impression for someone that doesn't have as much context as me.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+
+Lets look at the following security rules that setup a "restricted" and a "free" role as an example:
```
AlexicaWright:
```suggestion
Let's look at the following security rules that set up a "restricted" and a "free" role as an example:
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+
+     GRANT TRAVERSE ON GRAPH * NODES Person TO restricted;
+     DENY TRAVERSE ON GRAPH * NODES Customer TO restricted;
+     GRANT TRAVERSE ON GRAPH * ELEMENTS * TO free;
+
+Now, lets look at what the database needs to do in order to execute the following query:
```
AlexicaWright:
```suggestion
Now, let's look at what the database needs to do in order to execute the following query:
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
++--------------------------+
+| Operator                 |
++--------------------------+
+| +ProduceResults          |
+| |                        +
+| +NodeCountFromCountStore |
++--------------------------+
+....
+
+Internally however, very different operations need to be executed. The following table illustrate the difference.
```
AlexicaWright:
```suggestion
Internally however, very different operations need to be executed. The following table illustrates the difference.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+[%header,cols=2*]
+|===
+|User with "free" role
+|User with "restricted" role
+
+|The database can access the count store and retrieve the total number of nodes with the label `:Person`.
+
+This is a very quick operation.
+
+|The database can not just access the count store because it must make sure that only nodes with the desired label `:Person` are counted.
```
AlexicaWright:
```suggestion
|The database cannot just access the count store because it must make sure that only nodes with the desired label `:Person` are counted.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+This means that if the security model causes fewer results to be returned due to restricted read access in <<administration-security-subgraph, Graph and sub-graph access control>>,
+the index will also return the same fewer results.
+
+However, this rule is not correctly obeyed by <<administration-indexes-fulltext-search, Indexes for full-text search>>.
+These specific indexes are backed by Lucene internally.
+It is therefor not possible to know for certain whether a security violation occured for each specific entry returned from the index.
+As a result, Neo4j will return zero results from Full-text indexes if it is determined that any result might violate the security privileges active for that query.
+
+Since full-text indexes are not automatically used by Cypher, this does not lead to the case where the same Cypher query would return different results simple because such an index got created.
```
Lojjs: ... return different results **simply** because ... Also if we ever implement Cypher support for fulltext indexes, we should keep this in mind

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+Do to this, each node with the `:Person` label needs to be accessed and examined to make sure that it does not also have any blacklisted label like `:Customer`.
```
AlexicaWright:
```suggestion
Due to this, each node with the `:Person` label needs to be accessed and examined to make sure that it does not also have a denylisted label, such as `:Customer`.
```
Also, where was the label `:Customer` denylisted (blacklisted, see other comment?

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
+As described in <<administration-indexes-search-performance, Indexes for search performance>>, Neo4j {neo4j-version} supports the creation and use of indexes to improve the performance of Cypher queries.
+The Neo4j security model will impact the results of queries that use the index.
+When using normal Neo4j indexes, a Cypher query will always return the same results it would have if no index existed.
+This means that if the security model causes fewer results to be returned due to restricted read access in <<administration-security-subgraph, Graph and sub-graph access control>>,
+the index will also return the same fewer results.
+
+However, this rule is not correctly obeyed by <<administration-indexes-fulltext-search, Indexes for full-text search>>.
```
AlexicaWright:
```suggestion
However, this rule is not fully obeyed by <<administration-indexes-fulltext-search, Indexes for full-text search>>.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+These specific indexes are backed by Lucene internally.
+It is therefor not possible to know for certain whether a security violation occured for each specific entry returned from the index.
```
AlexicaWright:
```suggestion
It is therefore not possible to know for certain whether a security violation occurred for each specific entry returned from the index.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
AlexicaWright:
```suggestion
Since full-text indexes are not automatically used by Cypher, this does not lead to the case where the same Cypher query would return different results simply because such an index got created.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+As a result, Neo4j will return zero results from Full-text indexes if it is determined that any result might violate the security privileges active for that query.
+
+Since full-text indexes are not automatically used by Cypher, this does not lead to the case where the same Cypher query would return different results simple because such an index got created.
+Users need to explicitly call procedures to use these indexes.
+The problem is only that if this behaviour is not understood by the user, they might expect the full text index to return results that a different but semantically similar Cypher query does.
```
AlexicaWright:
```suggestion
The problem is only that if this behavior is not understood by the user, they might expect the full text index to return results that a different, but semantically similar, Cypher query does.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
+As described in <<administration-indexes-search-performance, Indexes for search performance>>, Neo4j {neo4j-version} supports the creation and use of indexes to improve the performance of Cypher queries.
+The Neo4j security model will impact the results of queries that use the index.
+When using normal Neo4j indexes, a Cypher query will always return the same results it would have if no index existed.
+This means that if the security model causes fewer results to be returned due to restricted read access in <<administration-security-subgraph, Graph and sub-graph access control>>,
+the index will also return the same fewer results.
+
+However, this rule is not correctly obeyed by <<administration-indexes-fulltext-search, Indexes for full-text search>>.
+These specific indexes are backed by Lucene internally.
+It is therefor not possible to know for certain whether a security violation occured for each specific entry returned from the index.
+As a result, Neo4j will return zero results from Full-text indexes if it is determined that any result might violate the security privileges active for that query.
```
Lojjs: Here you are spelling `Full-text` with capital F, everywhere else it is spelled `full-text`

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
Lojjs: `return results` --> `return the same results`

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+Since full-text indexes are not automatically used by Cypher, this does not lead to the case where the same Cypher query would return different results simple because such an index got created.
+Users need to explicitly call procedures to use these indexes.
+The problem is only that if this behaviour is not understood by the user, they might expect the full text index to return results that a different but semantically similar Cypher query does.
+
+==== Example with blacklisted properties
+
+Consider the following example.
+The database has nodes with Label `:User` and these have properties `name` and `email`.
```
Lojjs: `Label` should probably be lowercase, `label`

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+We have a normal index on both properties:
```
Lojjs: Again, I'm a bit unsure on the usage of the word `normal` here. Maybe "We have one built-in index and one full-text index on both properties". Or maybe `native` instead of `built-in`, not sure which word we normally use.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+This query will operate exactly as before, returning the same single result, because nothing in this query relates to the blacklisted property.
+
+But consider the full-text index query:
+
+    CALL db.index.fulltext.queryNodes("userNames", "ndy") YIELD node, score
+    RETURN node.name
+
+The problem now is that we do not know if the results provided by the index were because of a match to the `name` or the `surname` property.
+Results from the surname need to be excluded by the security rules, because they require that the user cannot see any `surname properties.
```
Lojjs: the last `surname` here is only half escaped, leading to weird rendering

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+However, the security model is not able to introspect the _Lucene_ query to know what it will actually do, whether it works only on the allowed `name` property, or also on the disallowed `surname` property.
+We know that the earlier query returned a match for `Joe Andy` which should now be filtered out.
+So, in order to never result results the user should not be able to see, we have to block all results.
```
Lojjs: "result results" --> "return results"

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+and thus they do not exhibit the behavior this section aims to clarify.
+While this document will not mention relationships further, the general function of the traverse privilege also applies to them.
+
+For any node that is traversable, due to `GRANT TRAVERSE` or `GRANT MATCH`,
+the user can get information about the labels attached to the node by calling the built-in `labels()` function.
+In the case of nodes with multiple labels, this can seemingly result in labels being returned to which the user
+wasn't directly granted access to.
+
+To give an illustrative example, imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
+We also have user with the role "custom" as defined by:
```
Lojjs: Maybe the role name "custom" should be backticked instead of in quotes?

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+In contrast, executing
+
+    MATCH (n:B) RETURN n, labels(n);
+
+will return only the one node that has both labels: `:A :B`. Even though `:B` was not whitelisted for traversal, there is one
+node with that label accessible in the data because of the whitelisted label `:A` that is attached to the same node.
+
+// Get A,B but was only granted A
+
+If a user is denied traverse on a label they will never get results from any node that has this label (also)
```
Lojjs: I think the `(also)` here doesn't provide any value, it just make it harder to read.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+In contrast to the normal graph traversal described in the previous section the built-in `db.labels()` procedure
+is not processing the data graph itself but the security rules defined on the system graph.
+That means:
+
+* if a label is explicitly whitelisted (granted), it will be returned by this procedure.
+* if a label is blacklisted (denied) or isn't explicitly whitelisted it will not be returned by this procedure.
+
+To reuse the example of the previous section: imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
+We also have user with the role "custom" as defined by:
```
Lojjs: Maybe the role name "custom" should be backticked instead of in quotes?

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
Lojjs: Maybe the role names "restricted" and "free" should be backticked instead of in quotes? Same in the table header below

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
Lojjs: Maybe `only traversable nodes` or something like that?

Lojjs (changes requested):
The content is great! A bunch of smaller comments on language/style and the headlines

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+Users need to explicitly call procedures to use these indexes.
+The problem is only that if this behaviour is not understood by the user, they might expect the full text index to return results that a different but semantically similar Cypher query does.
+
+==== Example with blacklisted properties
+
+Consider the following example.
+The database has nodes with Label `:User` and these have properties `name` and `email`.
+We have a normal index on both properties:
+
+    CREATE INDEX FOR (n:User) FOR (n.name, n.surname);
```
AlexicaWright: I would suggest a code block for examples.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+    CALL db.index.fulltext.createNodeIndex("userNames",["User", "Person"],["name", "surname"]);
+
+[NOTE]
+Full-text indexes also support multiple labels.
+See <<administration-indexes-fulltext-search, Indexes for full-text search>> for more details on creating and using full-text indexes.
+
+After creating these indexes, it would appear we have two indexes accomplishing the same thing.
+However this is not completely accurate.
```
AlexicaWright:
```suggestion
However, this is not completely accurate.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+These two indexes behave in different ways and are focused on different use cases.
+A key difference is that full-text indexes are backed by Lucene, and will use the Lucene syntax for querying the index.
+
+This has consequences for users restricted on the labels or properties involved in the indexes.
+Ideally if the labels and properties in the index are blacklisted, we can correctly return zero results from both native indexes and full-text indexes.
+However, there exist edge cases where this is not so simple.
+
+Imagine the following nodes were added to the database:
+
```
AlexicaWright, line 42: As above, I suggest using a code block for examples.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+Full-text indexes also support multiple labels.
+See <<administration-indexes-fulltext-search, Indexes for full-text search>> for more details on creating and using full-text indexes.
+
+After creating these indexes, it would appear we have two indexes accomplishing the same thing.
+However this is not completely accurate.
+These two indexes behave in different ways and are focused on different use cases.
+A key difference is that full-text indexes are backed by Lucene, and will use the Lucene syntax for querying the index.
+
+This has consequences for users restricted on the labels or properties involved in the indexes.
+Ideally if the labels and properties in the index are blacklisted, we can correctly return zero results from both native indexes and full-text indexes.
```
AlexicaWright:
```suggestion
Ideally, if the labels and properties in the index are denylisted, we can correctly return zero results from both native indexes and full-text indexes.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+However, there exist edge cases where this is not so simple.
```
AlexicaWright:
```suggestion
However, there are borderline cases where this is not as simple.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+Imagine the following nodes were added to the database:
+
+    CREATE (:User {name:'Mark', surname:'Andy'});
+    CREATE (:User {name:'Andy', surname:'Anderson'});
+    CREATE (:User:Person {name:'Mandy', surname:'Smith'});
+    CREATE (:User:Person {name:'Joe', surname:'Andy'});
+
+Consider blacklisting on the label `:Person`.
+
+    DENY TRAVERSE Person ON GRAPH * TO users;
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+If the user runs a query that will use the native index:
+
+    MATCH (n:User) WHERE n.name CONTAINS 'ndy' RETURN n.name;
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+The problem now is that we do not know if the results provided by the index were because of a match to the `name` or the `surname` property.
+The steps taken by the query engine would be:
+
+* run a _Lucene_ query on the full-text index to produce results containing `ndy` in either property, leading to four results.
+* filter the results to exclude nodes that also have the label `:Person`, filtering out `Mandy` and `Joe` so we have two results.
+
+This difference in results is due to the `OR` relationship between the two properties in the index creation.
+
+==== Blacklisting properties
```
AlexicaWright: We're steering away from using the words "blacklist" and "whitelist", so I'd suggest either to say "Denylisting properties" or "Denying access on properties"

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+Now consider blacklisting properties, like the `surname` property:
```
AlexicaWright: As above.
```suggestion
Now consider denylisting/denying access on properties, like the `surname` property:
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+    DENY READ {surname} ON GRAPH * TO users;
+
+Now we run the same queries again:
+
+    MATCH (n:User) WHERE n.name CONTAINS 'ndy' RETURN n.name;
+
+This query will operate exactly as before, returning the same single result, because nothing in this query relates to the blacklisted property.
```
AlexicaWright: As above.
```suggestion
This query will operate exactly as before, returning the same single result, because nothing in this query relates to the denylisted property.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+    RETURN node.name
+
+The problem now is that we do not know if the results provided by the index were because of a match to the `name` or the `surname` property.
+Results from the surname need to be excluded by the security rules, because they require that the user cannot see any `surname properties.
+However, the security model is not able to introspect the _Lucene_ query to know what it will actually do, whether it works only on the allowed `name` property, or also on the disallowed `surname` property.
+We know that the earlier query returned a match for `Joe Andy` which should now be filtered out.
+So, in order to never result results the user should not be able to see, we have to block all results.
+The steps taken by the query engine would be:
+
+* Determine if the full-text index includes blacklisted properties
```
AlexicaWright: As above
```suggestion
* Determine if the full-text index includes denylisted properties
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+will now return no nodes.
+
+==== The db.labels() procedure
+
+In contrast to the normal graph traversal described in the previous section the built-in `db.labels()` procedure
+is not processing the data graph itself but the security rules defined on the system graph.
+That means:
+
+* if a label is explicitly whitelisted (granted), it will be returned by this procedure.
+* if a label is blacklisted (denied) or isn't explicitly whitelisted it will not be returned by this procedure.
```
AlexicaWright: As above
```suggestion
* if a label is denied or isn't explicitly allowed it will not be returned by this procedure.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+However, this rule is not correctly obeyed by <<administration-indexes-fulltext-search, Indexes for full-text search>>.
+These specific indexes are backed by Lucene internally.
+It is therefor not possible to know for certain whether a security violation occured for each specific entry returned from the index.
+As a result, Neo4j will return zero results from Full-text indexes if it is determined that any result might violate the security privileges active for that query.
+
+Since full-text indexes are not automatically used by Cypher, this does not lead to the case where the same Cypher query would return different results simple because such an index got created.
+Users need to explicitly call procedures to use these indexes.
+The problem is only that if this behaviour is not understood by the user, they might expect the full text index to return results that a different but semantically similar Cypher query does.
+
+==== Example with blacklisted properties
```
AlexicaWright: As above
```suggestion
==== Example with denylisted properties
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+However, there exist edge cases where this is not so simple.
+
+Imagine the following nodes were added to the database:
+
+    CREATE (:User {name:'Mark', surname:'Andy'});
+    CREATE (:User {name:'Andy', surname:'Anderson'});
+    CREATE (:User:Person {name:'Mandy', surname:'Smith'});
+    CREATE (:User:Person {name:'Joe', surname:'Andy'});
+
+Consider blacklisting on the label `:Person`.
```
AlexicaWright:
```suggestion
Consider denylisting on the label `:Person`.
```
alternatively, "Consider denying access ..."

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+    MATCH (n:A) RETURN n, labels(n);
+
+they would be returned two nodes: the node that was labeled with `:A` and the node with labels `:A :B`.
+
+In contrast, executing
+
+    MATCH (n:B) RETURN n, labels(n);
+
+will return only the one node that has both labels: `:A :B`. Even though `:B` was not whitelisted for traversal, there is one
```
AlexicaWright: See other comment about black/whitelisting
```suggestion
will return only the one node that has both labels: `:A :B`. Even though `:B` was not allowlisted for traversal, there is one
```
Alternatively, "..was not allowed access for traversal..."

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+node with that label accessible in the data because of the whitelisted label `:A` that is attached to the same node.
```
AlexicaWright: As above
```suggestion
node with that label accessible in the data because of the allowlisted label `:A` that is attached to the same node.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+* if a label is explicitly whitelisted (granted), it will be returned by this procedure.
+* if a label is blacklisted (denied) or isn't explicitly whitelisted it will not be returned by this procedure.
+
+To reuse the example of the previous section: imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
+We also have user with the role "custom" as defined by:
+
+    GRANT TRAVERSE ON GRAPH * NODES A TO custom;
+
+This means that only label `:A` is explicitly whitelisted.
```
AlexicaWright:
```suggestion
This means that only label `:A` is explicitly allowlisted.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+
+Now we run the same queries again:
+
+    MATCH (n:User) WHERE n.name CONTAINS 'ndy' RETURN n.name;
+
+This query will operate exactly as before, returning the same single result, because nothing in this query relates to the blacklisted property.
+
+But consider the full-text index query:
+
+    CALL db.index.fulltext.queryNodes("userNames", "ndy") YIELD node, score
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
AlexicaWright:
```suggestion
So, in order to never return results the user should not be able to see, we have to block all results.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,101 @@
...
+Results from the surname need to be excluded by the security rules, because they require that the user cannot see any `surname properties.
+However, the security model is not able to introspect the _Lucene_ query to know what it will actually do, whether it works only on the allowed `name` property, or also on the disallowed `surname` property.
+We know that the earlier query returned a match for `Joe Andy` which should now be filtered out.
+So, in order to never result results the user should not be able to see, we have to block all results.
+The steps taken by the query engine would be:
+
+* Determine if the full-text index includes blacklisted properties
+* If yes, return an empty results stream, otherwise process as before
+
+The query will therefor return zero results in this case, rather than simply returning only the `Andy` result that might be expected.
```
AlexicaWright:
```suggestion
The query will therefore return zero results in this case, rather than simply returning only the `Andy` result that might be expected.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
+=== How does role-based access control impact label visibility
+
+==== Traversing the graph with multi-labeled nodes
+
+The general influence of access control privileges on graph traversal is described in detail in <<administration-security-subgraph, Graph and sub-graph access control>>.
+The following section will only focus on nodes because of their ability to have multiple labels. Relationships can only ever have one type
+and thus they do not exhibit the behavior this section aims to clarify.
+While this document will not mention relationships further, the general function of the traverse privilege also applies to them.
```
AlexicaWright:
```suggestion
While this section will not mention relationships further, the general function of the traverse privilege also applies to them.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+For any node that is traversable, due to `GRANT TRAVERSE` or `GRANT MATCH`,
+the user can get information about the labels attached to the node by calling the built-in `labels()` function.
+In the case of nodes with multiple labels, this can seemingly result in labels being returned to which the user
+wasn't directly granted access to.
+
+To give an illustrative example, imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
+We also have user with the role "custom" as defined by:
+
+    GRANT TRAVERSE ON GRAPH * NODES A TO custom;
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+If that user were to execute
+
+    MATCH (n:A) RETURN n, labels(n);
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+they would be returned two nodes: the node that was labeled with `:A` and the node with labels `:A :B`.
+
+In contrast, executing
+
+    MATCH (n:B) RETURN n, labels(n);
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+will return only the one node that has both labels: `:A :B`. Even though `:B` was not whitelisted for traversal, there is one
+node with that label accessible in the data because of the whitelisted label `:A` that is attached to the same node.
+
+// Get A,B but was only granted A
+
+If a user is denied traverse on a label they will never get results from any node that has this label (also)
+attached to it. Thus, the label name will never show up for them. For our example this can be done by executing:
+
+    DENY TRAVERSE ON GRAPH * NODES B TO custom;
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+The query
+
+    MATCH (n:A) RETURN n, labels(n);
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+will now return the node only labeled with `:A`, while the query
+
+    MATCH (n:B) RETURN n, labels(n);
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+will now return no nodes.
+
+==== The db.labels() procedure
+
+In contrast to the normal graph traversal described in the previous section the built-in `db.labels()` procedure
```
AlexicaWright:
```suggestion
In contrast to the normal graph traversal described in the previous section, the built-in `db.labels()` procedure
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+is not processing the data graph itself but the security rules defined on the system graph.
+That means:
+
+* if a label is explicitly whitelisted (granted), it will be returned by this procedure.
+* if a label is blacklisted (denied) or isn't explicitly whitelisted it will not be returned by this procedure.
+
+To reuse the example of the previous section: imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
+We also have user with the role "custom" as defined by:
+
+    GRANT TRAVERSE ON GRAPH * NODES A TO custom;
```
AlexicaWright: Code block

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -0,0 +1,69 @@
...
+
+This means that only label `:A` is explicitly whitelisted.
+Thus, executing
+
+    CALL db.labels();
```
AlexicaWright: Code block

AlexicaWright (commented):
Apart from the comments, please verify that everything appears where it should when built. When I build it locally, I get the section on "Known limitations of security" right below the index of chapter 5.5 "Security". 
Overall, it is a very good addition, my comments are mostly editorial.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-db-operations.asciidoc
SaschaPeukert: Line 8:
DENY TRAVERSE ON GRAPH * NODES Customer TO restricted;

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
SaschaPeukert:
```suggestion
will return only the one node that has both labels: `:A :B`. Even though `:B` was not allowed access for traversal, there is one
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -73,15 +73,15 @@ This difference in results is due to the `OR` relationship between the two prope
 
 ==== Blacklisting properties
 
-Now consider blacklisting properties, like the `surname` property:
+Now consider denylisting/denying access on properties, like the `surname` property:
```
AlexicaWright: Here I meant for you to select one of the options, either _denylisting_ **or** _denying access on_.

AlexicaWright (changes requested):
Looks great @SaschaPeukert ! Just one last comment to address.

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -13,7 +11,7 @@ In the case of nodes with multiple labels, this can seemingly result in labels b
 wasn't directly granted access to.
 
 To give an illustrative example, imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
-We also have user with the role "custom" as defined by:
+We also have user with a role `custom` as defined by:
```
Lojjs:
```suggestion
We also have a user with a role `custom` as defined by:
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-labels.asciidoc
```diff
@@ -73,7 +71,7 @@ That means:
 * if a label is denied or isn't explicitly allowed it will not be returned by this procedure.
 
 To reuse the example of the previous section: imagine a graph with three nodes: one labeled `:A`, one labeled `:B` and one with `:A :B`.
-We also have user with the role "custom" as defined by:
+We also have user with a role `custom` as defined by:
```
Lojjs:
```suggestion
We also have a user with a role `custom` as defined by:
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,125 @@
...
+
+The problem now is that we do not know if the results provided by the index were because of a match to the `name` or the `surname` property.
+The steps taken by the query engine would be:
+
+* run a _Lucene_ query on the full-text index to produce results containing `ndy` in either property, leading to four results.
+* filter the results to exclude nodes that also have the label `:Person`, filtering out `Mandy` and `Joe` so we have two results.
+
+This difference in results is due to the `OR` relationship between the two properties in the index creation.
+
+==== Blacklisting properties
```
Lojjs: Yo missed to update this one to denylisting

cypher/cypher-docs/src/docs/dev/ql/administration/security/security-and-indexes-intro.asciidoc
```diff
@@ -0,0 +1,125 @@
...
+But consider the full-text index query:
+
+[source, cypher]
+----
+CALL db.index.fulltext.queryNodes("userNames", "ndy") YIELD node, score
+RETURN node.name
+----
+
+The problem now is that we do not know if the results provided by the index were because of a match to the `name` or the `surname` property.
+Results from the surname need to be excluded by the security rules, because they require that the user cannot see any `surname properties.
```
Lojjs:
```suggestion
Results from the surname need to be excluded by the security rules, because they require that the user cannot see any `surname` properties.
```

Lojjs: This still looks wrong when I build

Lojjs (commented):
I found 2 new errors and 2 old ones

Lojjs (approved):

PR #647, opened 2020-08-28
Replace usage of sub-graph with path

davidoliverSP2:
Suggested replacement of `sub-graphs` with `paths`, via @petraselmer 

> The term "sub-graphs", whilst technically correct (in that a path is a simple form of a sub-graph), is highly misleading. `size(pattern)` is forbidden in the context of multiple path patterns, so referring to `sub-graphs` could lead to users trying to use `size` erroneously. The explanatory text correctly refers to paths only.


Lojjs (approved):
I agree that sub-graphs could be misleading, paths feels like a better option

PR #649, opened 2020-09-02
And RETURN examples to SHOW USERS/PRIVILEGES/DATABASES

mnd999:
Adds examples to the docs for the RETURN clause on the SHOW commands. I'm not sure how much to add for this so any feedback is welcome.

Depends on https://github.com/neo-technology/neo4j/pull/6955

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/security/role-management-syntax.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/security/role-management-syntax.asciidoc
@@ -6,9 +6,9 @@
 | [source, cypher]
 ----
 SHOW [ALL\|POPULATED] ROLES
-    [YIELD { * \| field1[, field2] } [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
+    [YIELD { * \| field1[, ...] } [ORDER BY field1 [, ...]] [SKIP n] [LIMIT n]]
     [WHERE expression]
-    [RETURN field1[, field2] [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
+    [RETURN field1[, ...] [ORDER BY field1 [, ...]] [SKIP n] [LIMIT n]]
 ----
 | List roles.
 | <<administration-security-administration-dbms-privileges-role-management, SHOW ROLE>>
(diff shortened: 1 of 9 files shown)


davidoliverSP2:
Thanks @mnd999 - could you tag which PR this depends on? many thanks

cypher/cypher-docs/src/docs/dev/ql/administration/security/role-management-syntax.asciidoc
```diff
@@ -6,8 +6,9 @@
 | [source, cypher]
 ----
 SHOW [ALL\|POPULATED] ROLES
-    [YIELD field1[, field2] [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
+    [YIELD { * \| field1[, field2] } [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
```
Lojjs: In the security documentation, we do `[, ...]` instead of ` [, field2]` to mark that it could be one or several values. I don't have a super strong opinion on this but if you read it literally I guess this way of writing it would suggest that you can have either 1 or 2 values but not 3.

cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
```diff
@@ -6,8 +6,9 @@
 | [source, cypher]
 ----
 SHOW USERS
-    [YIELD field1[, field2] [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
+    { * \| field1[, field2] } [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
```
Lojjs: This line seems to miss a `[YIELD ` first

cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
```diff
@@ -17,6 +18,9 @@ SHOW USERS
 | [source, cypher]
 ----
 SHOW USER[S] [name1[, name2]] PRIVILEGES
+    { * \| field1[, field2] } [ORDER BY field1 [, field2]] [SKIP n] [LIMIT n]]
```
Lojjs: This line seems to miss a `[YIELD ` first

Lojjs (changes requested):
A few minor comments. I think it is an ok level of effort but maybe we should have one example with RETURN with ORDER BY and SKIP/LIMIT as well?

Lojjs (approved):
I haven't built it but based on the PR it looks good now

davidoliverSP2 (approved):
Thanks @mnd999 & @Lojjs 👍 

I've pushed a couple of small editorial fixes, and will merge when green.

PR #661, opened 2020-09-14
Make sure nodes are connected to the tx in which they are used.

michael-simons:
That should go into 4.0, 4.1, 4.2 regarding @LinneaAndersson's message in product engineering.


LinneaAndersson (approved):
LGTM 👍

PR #682, opened 2020-10-05
updated built-in roles to include traverse

AlexicaWright:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SecurityUserAndRoleManagementTest.scala
```diff
@@ -160,8 +160,8 @@ class SecurityUserAndRoleManagementTest extends DocumentingTest with QueryStatis
             """This is the same command as `SHOW ALL ROLES`.
               |When first starting a Neo4j DBMS there are a number of built-in roles:
               |
-              |* `reader` - can perform read-only queries on all databases except `system`.
-              |* `editor` - can perform read and write operations on all databases except `system`, but cannot make new labels or relationship types.
+              |* `reader` - can traverse and perform read-only queries on all databases except `system`.
+              |* `editor` - can traverse and perform read and write operations on all databases except `system`, but cannot make new labels or relationship types.
```
Hunterness: Not sure what I think about this, not particularly like it though. Maybe something like:
```suggestion
              |* `reader` - can perform queries that only traverse and read on all databases except `system`.
              |* `editor` - can perform traverse, read and write operations on all databases except `system`, but cannot make new labels or relationship types.
```
or
```suggestion
              |* `reader` - can perform traverse and read operations on all databases except `system`.
              |* `editor` - can perform traverse, read and write operations on all databases except `system`, but cannot make new labels or relationship types.
```
instead?

PR #667, opened 2020-09-16
Add documentation for the new `EXECUTE` privileges

Hunterness:
Covers `EXECUTE PROCEDURE`, `EXECUTE BOOSTED PROCEDURE`, `EXECUTE ADMIN PROCEDURES`, `EXECUTE FUNCTION` and `EXECUTE BOOSTED FUNCTION`.

Depends on https://github.com/neo-technology/neo4j/pull/7008 and https://github.com/neo-technology/neo4j/pull/7322

PR #691, opened 2020-10-12
Remove merge conflicts introduced in #685

davidoliverSP2:
@AlexicaWright A quick fix - can you just check I removed the right part?! Presumably we need to retain the PUBLIC role definition.

Many thanks


AlexicaWright (approved):
Yes, you got the right part to keep! Thank you @davidoliverSP2 ! How did I miss this?

davidoliverSP2:
Resolved in #667

PR #693, opened 2020-10-13
Clarify why some metrics don't always appear in plans

pontusmelke:


davidoliverSP2 (approved):
Thanks @pontusmelke 👍

PR #702, opened 2020-10-20
Make the S in PRIVILEGES optional for `SHOW PRIVILEGES`

Hunterness:
Just updates the full syntax tables, don't think we need any other updates


hedengran (approved):

davidoliverSP2 (approved):
👍  thanks @Hunterness

PR #707, opened 2020-10-27
Admonition about the naming convention for creating databases and roles

renetapopova:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/DatabasesTest.scala
```diff
@@ -67,6 +60,18 @@ class DatabasesTest extends DocumentingTest with QueryStatisticsTestSupport {
...
+      note {
+        p("""Database names are subject to the <<cypher-manual#cypher-naming, standard Cypher restrictions on valid identifiers>>.
+          |The following naming rules apply:""")
+        p(
+          """
+            |* Database name length must be between 3 and 63 characters.
+            |* The first character must be an ASCII alphabetic character.
+            |* Subsequent characters can be ASCII alphabetic (`mydatabase`), numeric characters (`mydatabase2`), dots (`main.db`), and dashes (enclosed within backticks, e.g., `CREATE DATABASE ++`main-db`++`).
+            |* Names cannot end with dots or dashes.
+            |* Names that begin with an underscore and with the prefix `system` are reserved for internal use.
```
davidoliverSP2: is the use of "and" here correct?

should it be:
>Names that begin with an underscore, or with the prefix `system`, are reserved for internal use.

davidoliverSP2 (approved):
Thanks @renetapopova 

Just one question to consider.

PR #714, opened 2020-10-28
fixed so that the default roles will be generated (#706)

martin-neotech:
The enterprise edition have role-based access control.


renetapopova (approved):
LGTM, thanks @martin-neotech!

PR #719, opened 2020-11-02
Switch the index and constraint descriptions

davidoliverSP2:


AlexicaWright (approved):
Yes, this makes perfect sense! Thank you @davidoliverSP2 !

PR #720, opened 2020-11-03
split content across multiple files

recrwplay:
- Moves operators summary and dbhits to new topics
- Removes the redundant list from the index topic
- Introduction no longer needs to be a heading
- Operators topic has a page attribute set so the table of contents on the right hand side is not displayed
- Changes the column widths in the Operators Summary topic


davidoliverSP2 (changes requested):
thanks @recrwplay 

I'm not sure this is working as anticipated just yet....

There is a mismatch in the build TOC, and the order of pages in the submenu. Happy to take a look at the solution with you if that's helpful?

Screenshot 2020-11-18 at 13 41 38

recrwplay:
I couldn't figure out why this didn't behave as I was expecting, but if you build neo4j-documentation with this PR (or just download the jar files from the TC build) and the latest neo4j-manual-modeling 4.1 you should see the correct structure. There should be no list in this topic, and Introduction is no longer a heading

Screenshot 2020-11-19 at 14 24 09

davidoliverSP2:
I'll give it a go

davidoliverSP2 (approved):

davidoliverSP2:
Nicely done - and I rebuilt this locally by cherry picking the jars from the TC build.

👍

PR #739, opened 2020-11-13
Update glossary and prettifier parser after OPTIONS

Hunterness:
Was missed when `OPTIONS` were added to `CREATE INDEX`, `CREATE CONSTRAINT ... IS UNIQUE` and `CREATE CONSTRAINT ... IS NODE KEY`


Lojjs (approved):
Apart from the separate rendering bug of the glossary that @recrwplay is already tracking, this looks good.

AlexicaWright (approved):
LGTM thank you @Hunterness !

cypher/cypher-prettifier/src/test/scala/org/neo4j/cypher/docgen/tooling/PrettifierParserTest.scala
```diff
@@ -82,23 +105,31 @@ class PrettifierParserTest extends ParserTestBase[Seq[SyntaxToken], Seq[SyntaxTo
     }
   }
 
-  test("shouldParseDropConstraintByName") {
```
Hunterness, line 85: Moved this test up above `shouldParseShowConstraints` to be consistent with the index tests, it has confused github

PR #742, opened 2020-11-16
reworked the path uniqueness page

martin-neotech:


cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
 [[cypher-result-uniqueness]]
-= Cypher Result Uniqueness
+= Cypher path matching
 
-While pattern matching, Neo4j Cypher makes sure to not include matches where the same graph relationship is found multiple times in a single path matching a particular pattern.
-This is called _relationship isomorphism_ and is a very effective way of reducing the result set size and preventing infinite traversals.
-To understand this better, let's consider a few alternative options for uniqueness:
+:description: Cypher path matching uses relationship isomorphis, the same relationship cannot be returned more than once in the same result record.
```
davidoliverSP2, line 4: I think this should be an abstract, rather than a description?
```suggestion
[abstract]
--
Cypher path matching uses relationship isomorphis, the same relationship cannot be returned more than once in the same result record. 
--
```

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+
+[NOTE]
+====
+In Neo4j all relationships have a direction.
+However, you can have the notion of undirected relationships at query time.
+====
+
+In the case of variable length pattern expressions, it is particularly important to have a constraint check, or an infinite number of result records could be found.
+
+To understand this better, let us consider a few alternative options:
```
davidoliverSP2, line 16: I think this sentence can be removed?

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+**Constraints:** No constraints for path matching.
+
+
+.Homomorphism
+====
+The graph is composed of only two nodes `(a)` and `(b)`, connected by one relationship, `(a)&#x2D;&#x2D;&#x3E;(b)`.
+
+If the query is looking for paths of length `n` and do not care about the direction, a path of length `n` will be returned repeating the two nodes over and over.
+
+For example, find all paths with 5 relationships and do not care about the relationship direction.
```
davidoliverSP2:
```suggestion
For example, find all paths with 5 relationships and do not care about the relationship direction:
```

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+
+[source, cypher]
+----
+MATCH p=()-[*5]-() RETURN nodes(p)
+----
+
+could return the two resulting records `[a,b,a,b,a,b]` as well as `[b,a,b,a,b,a]`.
```
davidoliverSP2: This sentence sounds vague - I would suggest something more confident:
```suggestion
This will return the two resulting records `[a,b,a,b,a,b]`, as well as `[b,a,b,a,b,a]`.
```

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+.Node isomorphism
+====
+The graph is composed of only two nodes `(a)` and `(b)`, connected by one relationship, `(a)&#x2D;&#x2D;&#x3E;(b)`.
+
+[source, cypher]
+----
+MATCH p=()-[*1]-() RETURN nodes(p)
+----
+
+could return the two resulting records `[a, b]` as well as `[b, a]`.
```
davidoliverSP2: Also vague, as the previous comment.

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+====
+
+
+== Relationship isomorphism
+
+**Constraints:** The same relationship cannot be returned more than once for each path matching record.
+
+In the above two-node example, `(a)&#x2D;&#x2D;&#x3E;(b)`, only paths of length 1 can be found with the relationship isomorphism constraint.
```
davidoliverSP2: As above:
```suggestion
In another two-node example, such as `(a)&#x2D;&#x2D;&#x3E;(b)`; only paths of length 1 can be found with the relationship isomorphism constraint:
```

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+
+could return the two resulting records `[a,b,a,b,a,b]` as well as `[b,a,b,a,b,a]`.
+====
+
+
+== Node isomorphism
+
+**Constraints:** The same node cannot be returned more than once for each path matching record.
+
+In the above two-node example, `(a)&#x2D;&#x2D;&#x3E;(b)`, only paths of length 1 can be found with the node isomorphism constraint.
```
davidoliverSP2: In which example? the Homomorphism example? I think we should try and be specific here - something like:
```suggestion
In another two-node example, such as `(a)&#x2D;&#x2D;&#x3E;(b)`, only paths of length 1 can be found with the node isomorphism constraint.
```

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,107 @@
...
+could return the two resulting records `[a, b]` as well as `[b, a]`.
+====
+
+
+== Cypher path matching examples
+
+Cypher makes use of relationship isomorphism for path matching.
+
+
+.Friend of friends
```
davidoliverSP2, line 86: I think this section is a little confusing, with the single example, and numbered sub-queries. I'm not sure of the solution though - I first thought the queries would be better as numbered examples.

However, I wonder whether the layout is better as-is, and just to reconsider the wording around the last query, which is a little confusing comparing back to Query 2.

davidoliverSP2 (changes requested):
Thanks @martin-neotech - A few comments to consider. 

Also, I noticed that the in page TOC on the _introduction/index_ page still lists *Result uniqueness* - I don't think you need to update that, but that whole TOC could simply be removed.

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
martin-neotech: I will leave it as it is. It is because the antora build uses the abstract to extract a description. So the new way forward should be to remove abstracts that look ugly on the pages and insert descriptions.

cypher/cypher-docs/src/docs/graphgists/uniqueness/uniqueness.asciidoc
```diff
@@ -1,34 +1,104 @@
 [[cypher-result-uniqueness]]
-= Cypher Result Uniqueness
+= Cypher path matching
 
-While pattern matching, Neo4j Cypher makes sure to not include matches where the same graph relationship is found multiple times in a single path matching a particular pattern.
-This is called _relationship isomorphism_ and is a very effective way of reducing the result set size and preventing infinite traversals.
-To understand this better, let's consider a few alternative options for uniqueness:
+:description: Cypher path matching uses relationship isomorphis, the same relationship cannot be returned more than once in the same result record.
```
craigtaverner, line 4: Isomorphism is spelled wrong. I also think that it would be a very good idea if changing this section to get feedback from the GQL or langstar people, as the nomenclature is being clarified and solidified a lot by the openCypher and GQL projects.

craigtaverner (changes requested):
One spelling mistake. I did not render the docs so do not know how good the rest looks, but did not see any other obvious mistakes.

PR #744, opened 2020-11-16
all relationships have a direction

martin-neotech:


cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
```diff
@@ -69,13 +69,26 @@ Structural types comprise:
    ** Id
```
renetapopova: Just a suggestion in order to avoid the repetition of the verb _comprise_, to change the "Structural types comprise:" to "Structural types are composed of:"

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
```diff
@@ -69,13 +69,26 @@ Structural types comprise:
...
+
+[NOTE]
+====
+Labels are not values but are a form of pattern syntax.
+====
 
-NOTE: Labels are not values but are a form of pattern syntax.
+[NOTE]
+====
+In Neo4j all relationships have a direction.
```
renetapopova:
```suggestion
In Neo4j, all relationships have a direction.
```

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
```diff
@@ -69,13 +69,26 @@ Structural types comprise:
...
+However, you can have the notion of undirected relationships at query time.
+====
```
renetapopova, line 88: I wonder if all three notes can be merged into one note.

renetapopova (commented):
Thanks @martin-neotech! I think it's better now, just minor suggestions.

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
martin-neotech: Comprised of is an expression in English that means "to compose or constitute". While its use is common in writing and speech, it has been regarded by some language professionals as incorrect, stemming from the fact that comprise on its own already means "composed of". - https://en.wikipedia.org/wiki/Comprised_of

I will then remove the word `comprised of` for the `composed of`

renetapopova: I don't remember seeing `comprised of`, just `comprise` and `comprising`.

martin-neotech: True, `comprise` was there, but I have replaced it with `composed of`.

AlexicaWright: My two cents here, there is a difference between _comprise_ and _compose_. It has to do with parts and the whole.  In this case, _comprise_ is the correct choice, as in _Structural types are comprised of_ alt _Structural types are comprises_

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
AlexicaWright: Agree, these could be merged into one note.

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
```diff
@@ -86,10 +99,10 @@ NOTE: Labels are not values but are a form of pattern syntax.
 * [ ] Cannot be stored as properties
 * [*] Can be constructed with <<cypher-expressions, Cypher literals>>
 
-Composite types comprise:
+Composite types are composed of:
```
AlexicaWright:
```suggestion
Composite types are comprised of:
```

cypher/cypher-docs/src/docs/dev/syntax/values.asciidoc
```diff
@@ -86,10 +99,10 @@ NOTE: Labels are not values but are a form of pattern syntax.
...
 
-* *Lists* are heterogeneous, ordered collections of values, each of which has any property, structural or composite type.
-* *Maps* are heterogeneous, unordered collections of (key, value) pairs, where:
+* **Lists** are heterogeneous, ordered collections of values, each of which has any property, structural or composite type.
```
AlexicaWright:
```suggestion
* **Lists** are heterogeneous, ordered collections of values, each of which has any property, structural, or composite type.
```

AlexicaWright (commented):
Great addition, thank you @martin-neotech ! A few comments :)

PR #746, opened 2020-11-16
The label/type list of entity is not optional and other small syntax updates

Hunterness:
* Backported 4.1 version of read/traverse/match syntax since it was correct there
* Also backported `"{"` for the property list in read/match to show that is part of the syntax
* For revoke commands, don't need both `[] `and `{}` to group optional parts

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-match-syntax.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-match-syntax.asciidoc
@@ -2,11 +2,12 @@
 [source, cypher]
 -----
 DENY MATCH
-    { * | properties }
+    "{" { * | properties } "}"
     ON GRAPH[S] {name | *}
         [
-            { NODE[S] | RELATIONSHIP[S] | ELEMENT[S] }
-            [ { * | labels } ]
+            ELEMENT[S] { * | label-or-rel-type-name[,...] }
+            | NODE[S] { * | label-name[,...] }
+            | RELATIONSHIP[S] { * | rel-type-name[,...] }
         ]
     TO roles
 -----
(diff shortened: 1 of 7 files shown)


OliviaYtterbrink (approved):

cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-match-syntax.asciidoc
```diff
@@ -2,11 +2,12 @@
 [source, cypher]
 -----
 DENY MATCH
-    { * | properties }
+    "{" { * | properties } "}"
```
Hunterness, line 5: @OliviaYtterbrink should this be `"{" { * | property[, ...] } "}"`? (probably a bit of everywhere and multiple versions so I guess I should make a card for it)

Hunterness: I guess same goes for `roles` -> `role[, ...]`

OliviaYtterbrink: Just make sure we actually support multiples of all of them in 4.0....

Hunterness: For properties and roles I'd say we do, don't think those were ever singular, but yes would be good to confirm (and see if there are more versions like this)

Hunterness: Added a card, https://trello.com/c/nZWT53fV/2265-graph-privileges-comma-separated-lists

renetapopova (approved):
Looks good! Thanks @Hunterness

PR #747, opened 2020-11-16
Add back `REVOKE ROLE role FROM user` in syntax table and other syntax updates

Hunterness:
* When privileges was added to the list it was missed from REVOKE ROLE which lead to that table entry not being rendered
* Fix so we print the neo4j version instead of printing `${neo4j.version}`
* Update transaction privileges in database privilege list to use correct `()` instead of `{}`
* Fix syntax for transaction privileges to show what is part of syntax and what is just grouping
* For revoke commands, don't need both `[]` and `{}` to group optional parts
* Don't print `\|` in syntax examples when it should just be `|`
* Update so `CREATE OR REPLACE` says replacing instead of just creating

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/security/database/admin-role-database.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/security/database/admin-role-database.asciidoc
@@ -22,9 +22,9 @@ The components of the database privilege commands are:
 ** `CREATE NEW [PROPERTY] NAME` - allows property names to be created, so that nodes and relationships can have properties with these names assigned.
 ** `NAME [MANAGEMENT]` - allows all of the name management capabilities: node labels, relationship types, and property names.
 ** `ALL [[DATABASE] PRIVILEGES]` - allows access, index, constraint, and name management for the specified database.
-** `SHOW TRANSACTION {* | user[, ...]}` -  allows listing transactions and queries for the specified users on the specified database.
-** `TERMINATE TRANSACTION {* | user[, ...]}` - allows ending transactions and queries for the specified users on the specified database.
-** `TRANSACTION [MANAGEMENT] {* | user[, ...]}` - allows listing and ending transactions and queries for the specified users on the specified database.
+** `SHOW TRANSACTION (* | user[, ...])` -  allows listing transactions and queries for the specified users on the specified database.
+** `TERMINATE TRANSACTION (* | user[, ...])` - allows ending transactions and queries for the specified users on the specified database.
+** `TRANSACTION [MANAGEMENT] (* | user[, ...])` - allows listing and ending transactions and queries for the specified users on the specified database.
 
 * _name_
 ** The database to associate the privilege with.
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/security/database/admin-role-database.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/security/database/admin-role-database.asciidoc
@@ -22,9 +22,9 @@ The components of the database privilege commands are:
 ** `CREATE NEW [PROPERTY] NAME` - allows property names to be created, so that nodes and relationships can have properties with these names assigned.
 ** `NAME [MANAGEMENT]` - allows all of the name management capabilities: node labels, relationship types, and property names.
 ** `ALL [[DATABASE] PRIVILEGES]` - allows access, index, constraint, and name management for the specified database.
-** `SHOW TRANSACTION (* | user[, ...])` -  allows listing transactions and queries for the specified users on the specified database.
-** `TERMINATE TRANSACTION (* | user[, ...])` - allows ending transactions and queries for the specified users on the specified database.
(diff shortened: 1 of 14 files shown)


cypher/cypher-docs/src/docs/dev/ql/administration/security/role-management-syntax.asciidoc
```diff
@@ -40,7 +40,7 @@ CREATE ROLE name [IF NOT EXISTS] [AS COPY OF name]
 ----
 CREATE OR REPLACE ROLE name [AS COPY OF name]
 ----
-| Create a new role.
+| Create a new role, replacing an existing role if one exists with the same name.
```
renetapopova: A suggestion for rewording:
"Create a new role, or if a role with the same name exists, replace it."

cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
```diff
@@ -43,7 +43,7 @@ CREATE OR REPLACE USER name
   [[SET PASSWORD] CHANGE [NOT] REQUIRED]
   [SET STATUS {ACTIVE \| SUSPENDED}]
 ----
-| Create a new user.
+| Create a new user, replacing an existing user if one exists with the same name.
```
renetapopova: same as above

renetapopova (commented):
Looks good, thanks @Hunterness! I have a minor suggestion for rewording.

cypher/cypher-docs/src/docs/dev/ql/administration/security/role-management-syntax.asciidoc
renetapopova: Or even: Create or replace a role with the specified name.

Hunterness: I updated this wording once from our code review, from `Recreate a role, or create a new role if no such role exists.` to `Create a new role, replacing an existing role if one exists with the same name.`

Think I prefer your first version
```suggestion
| Create a new role, or if a role with the same name exists, replace it.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
Hunterness:
```suggestion
| Create a new user, or if a user with the same name exists, replace it.
```

cypher/cypher-docs/src/docs/dev/ql/administration/security/database/admin-role-database.asciidoc
```diff
@@ -22,9 +22,9 @@ The components of the database privilege commands are:
...
 ** `NAME [MANAGEMENT]` - allows all of the name management capabilities: node labels, relationship types, and property names.
 ** `ALL [[DATABASE] PRIVILEGES]` - allows access, index, constraint, and name management for the specified database.
-** `SHOW TRANSACTION {* | user[, ...]}` -  allows listing transactions and queries for the specified users on the specified database.
-** `TERMINATE TRANSACTION {* | user[, ...]}` - allows ending transactions and queries for the specified users on the specified database.
-** `TRANSACTION [MANAGEMENT] {* | user[, ...]}` - allows listing and ending transactions and queries for the specified users on the specified database.
+** `SHOW TRANSACTION (* | user[, ...])` -  allows listing transactions and queries for the specified users on the specified database.
+** `TERMINATE TRANSACTION (* | user[, ...])` - allows ending transactions and queries for the specified users on the specified database.
+** `TRANSACTION [MANAGEMENT] (* | user[, ...])` - allows listing and ending transactions and queries for the specified users on the specified database.
```
Hunterness: Should probably be `({* | user[, ...]})`.

@OliviaYtterbrink we should probably decide how we want to do in these lists. For transaction it looked this way, for `SET/REMOVE LABEL` is just says `labels` and `READ/MATCH/SET PROPERTY` had `{props}`, neither which is fully correct either since they too can have `*` or a list (and property has `{}` as part of syntax while labels don't). Not sure if there are more places but we should decide how we want to have them in the lists (easiest is to just have the privilege name and not include the user/label/property list since they might get chunky, properties for example would be `"{" {* | prop[, ...]} "}"`)

Hunterness: `MERGE` privileges also have a property list or * in the syntax but not in the privilege list, there it only has the name.

Hunterness: See also my other comment about this, https://github.com/neo4j/neo4j-documentation/pull/762#discussion_r528561758

Hunterness:
@renetapopova Updated with your comments as well as some additional work from us (that I realized when looking at other parts of my recent docs work)

renetapopova (approved):
LGTM, thanks @Hunterness!

PR #755, opened 2020-11-18
Fix port clashes when running tests in parallel or alongside Neo4J

mnd999:
* Fix port clashes on online backup and rest server to allow tests to run alongside a neo4j server
* Update Scala version to match product


Hunterness (approved):
I think it looks good but I'm not overly familiar with ports :P

davidoliverSP2 (approved):
Thanks @mnd999 👍

PR #762, opened 2020-11-19
Update to `[, ...]` instead of plural name to indicate comma separated list

Hunterness:


Lojjs (approved):
Have not built the branch but only looking at the PR, everything looks good.

cypher/cypher-docs/src/docs/dev/ql/administration/security/grant-deny-syntax.asciidoc
```diff
@@ -9,23 +9,23 @@ The syntax is:
...
 | [source, cypher]
-REVOKE DENY graph-privilege ON GRAPH[S] {name \| *} [entity] FROM roles
+REVOKE DENY graph-privilege ON GRAPH[S] {name \| *} [entity] FROM role[, ...]
 | Revoke a denied privilege from one or multiple roles
 
 | [source, cypher]
-REVOKE graph-privilege ON GRAPH[S] {name \| *} [entity] FROM roles
+REVOKE graph-privilege ON GRAPH[S] {name \| *} [entity] FROM role[, ...]
 | Revoke a granted or denied privilege from one or multiple roles
 |===
```
Hunterness, line 31: Further down in this file is `{props}` in the privilege list (both for READ and MATCH), those should have been updated as well

Hunterness: there's also a `ELEMENTS+*+` which can be updated at the same time (shouldn't have `+`s)

davidoliverSP2:
@Hunterness We appear to have picked up merge conflicts in this PR - could you take a look?

cypher/cypher-docs/src/docs/dev/ql/administration/security/grant-deny-syntax.asciidoc
Hunterness: (if the `+` are removed confirm that the text doesn't get bold)

Hunterness:
@davidoliverSP2 I think i'm conflicting with my own PRs that have recently been merged, will take a look during the day :) (on this and other PRs)

cypher/cypher-docs/src/docs/dev/ql/administration/security/grant-deny-syntax.asciidoc
Hunterness: @Lojjs We should probably decide how we want to do in these lists, just adding `[, ...]` is not fully correct. A more accurate version would be `"{" {* | prop[, ...]} "}"`, which is very chunky.

We also have different handling of the user list for transaction privileges, the label list for label privileges and these property lists for read/match/set property and merge.

I think the easiest is to not include the users/labels/properties in these lists and just have the privilege names (this is what `MERGE` to)

See also my other comment about this, https://github.com/neo4j/neo4j-documentation/pull/747#discussion_r528557043

cypher/cypher-docs/src/docs/dev/ql/administration/security/grant-deny-syntax.asciidoc
```diff
@@ -66,7 +66,7 @@ this privilege can only be assigned to all nodes, relationships and properties i
 *** `ELEMENTS` label (both nodes and relationships).
 ** The label or type can be `+*+` which means all labels or types.
 ** Multiple labels or types can be specified, comma-separated.
-** Defaults to `ELEMENTS +*+` if omitted.
+** Defaults to `ELEMENTS` `+*+` if omitted.
```
Lojjs, line 69: Out of curiosity, can you explain what difference this leads to?

Hunterness: This is how it looks in 4.1 and forward, it's to not actually print the `+` but use them for escaping the `*` so you don't get bold text, it will however look like '`ELEMENTS` `*`' instead of '`ELEMENTS *`' (but at least don't have `ELEMENTS +*+`)

Not sure if it is needed in this case but I just followed how it had been fixed in 4.1

cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-match-syntax.asciidoc
```diff
@@ -2,12 +2,12 @@
 [source, cypher]
 -----
 DENY MATCH
-    "{" { * | properties } "}"
+    "{" { * | property[, ...] } "}"
```
AlexicaWright, line 5: Just out of curiosity, why is the syntax here [, ...], when it's [,...] (without a space separating the comma and the periods) on the other lines in the example? 
This applies to all of the `[, ...]` in this PR.

AlexicaWright (commented):
Just one comment.

cypher/cypher-docs/src/docs/dev/ql/administration/security/deny-match-syntax.asciidoc
Hunterness: I know that I in some of my PRs (not sure if this one or another) went over to try and make sure we had `thing[, ...]` instead of `thing [, ...]` but I guess I missed going over and checking `[, ...]` vs `[,...]` (and therefore probably also `thing [,...]`).

I know I prefer having the space`[, ...]` but I guess different people have added it differently which is why we now have a mix.

I guess I could (add a card to) go over and make sure its consistent

Hunterness: Since it looks like its on these PRs I might just go over and fix it here

Hunterness: Added an additional commit which updates so we should have `[, ...]` everywhere now :)

AlexicaWright (approved):
Thank you @Hunterness ! It looks so much better with a consistent use of `[, ...]`

PR #764, opened 2020-11-19
Updates around comma separated lists

Hunterness:
* `[, ...]` instead of plural name
* no space before `[, ...]`, ex. `name[, ...]` not `name [, ...]`
* update `grantee` to `role` to be consistent
* change `field1[, ...]` to `field[, ...]` to not have numbers, to be consistent


Lojjs (approved):
Have not built the branch but only looking at the PR, everything looks good.

cypher/cypher-docs/src/docs/dev/ql/administration/security/grant-deny-syntax.asciidoc
```diff
@@ -39,23 +39,23 @@ Graphs created after this command execution will also be associated with these p
 | Command | Description
```
Hunterness, line 40: Further up in this file we link to read and write privileges, in their list of privileges is `{props}`, those should have been updated as well

Hunterness: there's also a `ELEMENTS+*+` which can be updated at the same time (shouldn't have `+`s)

Hunterness: (if the `+` are removed confirm that the text doesn't get bold)

Hunterness: The `+` might have been fixed already but still looks slightly odd with a non-escaped space in between: `ELEMENTS` `*`

Hunterness: See, https://github.com/neo4j/neo4j-documentation/pull/762#discussion_r528561758

PR #768, opened 2020-11-20
4.3 cherry pick 734 and 739

AlexicaWright:


cypher/cypher-docs/src/docs/dev/ql/administration/constraints/index.asciidoc
```diff
@@ -254,7 +254,67 @@ include::drop-a-non-existing-constraint.asciidoc[leveloffset=+2]
 
 [[administration-constraints-list-constraint]]
 === List constraints
-include::list-constraints.asciidoc[leveloffset=+2]
+
+Listing all constraints can be done with `SHOW CONSTRAINTS`, which will produce a table with the following columns:
```
Hunterness: just realized that to be consistent with indexes we should have

```suggestion
Listing constraints can be done with `SHOW CONSTRAINTS`, which will produce a table with the following columns:
```

(also true for 4.2)

Hunterness:
Taken over by https://github.com/neo4j/neo4j-documentation/pull/779

PR #772, opened 2020-11-20
Escape the asterisks so they avoid the bold block (in the MATCH privilege section)

davidoliverSP2:
Resolves the unintended bold block here: https://neo4j.com/docs/cypher-manual/4.0/administration/security/subgraph/#administration-security-subgraph-match

@renetapopova assuming the build goes green, let's use this PR to pick the build artefacts and generate the antora output without having to locally compile the `neo4j-documentation` repo.


renetapopova (approved):
Looks good! Thanks @davidoliverSP2!

PR #779, opened 2020-11-25
4.3 cherry pick 734 and 739

Hunterness:
New version of https://github.com/neo4j/neo4j-documentation/pull/768 to make sure we got the correct commits and rebase correctly on 4.3


AlexicaWright (approved):
As far as I can tell, everything that should be in here, is in here. Thanks for sorting it out @Hunterness

PR #804, opened 2020-12-02
Add escape chars to work with pdf and html

recrwplay:
For html there was nothing wrong with the previous format, but the asciidoctor pdf back-end could not handle it. Using `+` to escape the characters inside the codeblocks produces the correct output for both html and pdf.


martin-neotech (approved):
Nice!

PR #809, opened 2020-12-03
More docs cleanup

Hunterness:

PR #810, opened 2020-12-03
More docs cleanup

Hunterness:


nadja-muller (approved):

renetapopova (approved):
Thanks @Hunterness. Same comment as in #809. I saw in https://neo4j.com/docs/cypher-manual/current/administration/indexes-for-search-performance/#administration-indexes-syntax that we also use `LabelName`, `propertyName`, `propertyName_1`, `propertyName_n`, in my opinion having `name` as a suffix, in whatever form, would be more self-explanatory than just for example `label`.

Hunterness:
We have slightly different namings of things between the index/constraint and administration commands, I tried to make sure we were consistent within index and constraint and then within the administration commands but decided not to make it consistent over both since they are separate.

I agree that having `name` might make it more self-explanatory but the administration commands are often quite long and having shorter names took precedence (and we used that in more places then not already)

PR #813, opened 2020-12-04
Update heading level

recrwplay:
The headings are at level 4 but expected to be level 3 at the point where they are included.


recrwplay:
[Address warnings in the Antora build](https://trello.com/c/Si0htooA/25-address-warnings-in-the-antora-build)

martin-neotech (approved):
Nice!

AlexicaWright (commented):
Builds just fine, but it renders no differently than the currently published docs?

recrwplay:
> Builds just fine, but it renders no differently than the currently published docs?

Yes, the html was forced to the correct header level by asciidoctor and a warning issued. This change just uses the correct level in the source so there is no warning in the build.

AlexicaWright:
Ah, I see! Then 👍  from me!

PR #825, opened 2020-12-11
More name clean-up

Hunterness:
Updating `label-name`, `rel-type-name` and `label-or-rel-type-name` to `label`, `rel-type` and `label-or-rel-type`


renetapopova (approved):
LGTM, thanks @Hunterness!

PR #832, opened 2020-12-16
cherry-picked (#720)

recrwplay:


recrwplay:
Actually build 3263 passed. Failed build is showing because the previous check was hanging in Team City.

PR #838, opened 2020-12-18
Improvements to syntax summaries section

Lojjs:


cypher/cypher-docs/src/docs/dev/ql/administration/security/index.asciidoc
```diff
@@ -77,15 +77,15 @@ These summaries will use some special characters to indicate such variations.
 
 The special characters and their meaning are as follows:
 
-.Special characters in syntax summeries
-[options="header", width="100%", cols="1a^,3a,3a"]
+.Special characters in syntax summaries
+[options="header", width="100%", cols="1a^,3a,3a,3"]
```
Hunterness: why the added 3?

Hunterness (approved):

cypher/cypher-docs/src/docs/dev/ql/administration/security/index.asciidoc
Lojjs: Without this number 3, everything was rendered into only 2 columns. Not sure if there is a better way to fix this, this was just trial-and-error from my side

Hunterness: it feels like you are specifying 4 columns but if you say so :P (Maybe the docs team knows why its needed :P)

recrwplay: According to [column specifiers](https://asciidoc.org/userguide.html#X70) in the Asciidoc user guide, the style (eg 'a' for 'asciidoc') should be specified last. The doc also claims that the order of align or width is not important, but I'm finding that it only works if I specify the alignment first in this table. So `cols="^1a,3a,3a"` works. (Also, `cols="^1a,2*3a"` would be valid).

Lojjs: @recrwplay Thanks, I will try using `cols="^1a,3a,3a"` istead, that looks more like the proper way to do it.

cypher/cypher-docs/src/docs/dev/ql/administration/security/index.asciidoc
```diff
@@ -106,7 +106,7 @@ It also groups alternatives together, when there can be either of the alternativ
 
 |===
 
-There are no other characters that need to be escaped.
+The special characters in the table above, are the only ones which needs to be escaped using `"` in the syntax summaries.
```
recrwplay:
```suggestion
The special characters in the table above are the only ones that need to be escaped using `"` in the syntax summaries.
```

recrwplay (approved):
Looks good!

PR #840, opened 2021-01-06
Fix operators comparison docs

recrwplay:
Move include statement to OperatorsTest because the include statement from the index page cannot work in the new build.


Lojjs (changes requested):
The operators page itself looks good now, but there is an additional menu on the syntax side https://neo4j.com/docs/cypher-manual/3.5/syntax/ which also needs updating. Right now it looks like this: 
Screenshot 2021-01-15 at 11 20 59

Lojjs (approved):
Looks good now!

PR #847, opened 2021-01-12
added filtering info to various refcards

AlexicaWright:


martin-neotech (approved):
Looks good, I will try to build it tomorrow.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
```diff
@@ -65,6 +65,17 @@ SHOW DATABASE myDatabase
...
+//
+
+SHOW DATABASES
+YIELD name, currentStatus
+WHERE name CONTAINS 'my' AND currentStatus = 'online'
+###
+
+List information about databases, filtered by name and online status and further refined by conditions on these.
+
+
```
Hunterness:
```suggestion

```

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
```diff
@@ -65,6 +65,17 @@ SHOW DATABASE myDatabase
 
 List information about the database `myDatabase`.
 
+###assertion=show-one
+//
+
+SHOW DATABASES
```
Hunterness: Where you added this we get the order:
```
SHOW DATABASES
SHOW DATABASE name
SHOW DATABASES YIELD ...
SHOW DEFAULT DATABASE
```
Not sure I like that, I'd say either move the new one last or second place (so either you list all different `SHOW` commands first and then the `YIELD`/`WHERE` or collect all `SHOW DATABASES` versions together before the other `SHOW` command versions)

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
```diff
@@ -65,6 +65,17 @@ SHOW DATABASE myDatabase
...
+YIELD name, currentStatus
+WHERE name CONTAINS 'my' AND currentStatus = 'online'
+###
+
+List information about databases, filtered by name and online status and further refined by conditions on these.
```
Hunterness: Not sure what I think of the use of `filtering` here, I guess we're filtering the columns so it should be fine but my brain still connects it to filtering rows which is what your `refined by` does :P (this goes for all commands not just database)

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
Hunterness: To follow the order used for `SHOW ROLES` it should be in second place:
```
SHOW DATABASES
SHOW DATABASES YIELD ...
SHOW DATABASE name
SHOW DEFAULT DATABASE
```

Hunterness (changes requested):

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
AlexicaWright: Updated!

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/DatabaseManagementTest.scala
AlexicaWright: I think that for a reader of the manual, the wording will be fine as `filtering` is not per definition associated with `columns`.

Hunterness (approved):

PR #831, opened 2020-12-15
Add missing section to administration toc

Hunterness:

PR #833, opened 2020-12-16
Move sentence to the right paragraph.

sherfert:

PR #860, opened 2021-01-19
Move sentence to the right paragraph

davidoliverSP2:
Resubmission of #831

For some reason, the previous PR had test issues, and wasn't generating output as expected. Perhaps it was a little too aged, and more recent changes had caused build problems. Resubmitting to save time in troubleshooting.


davidoliverSP2:
Approved on the original PR #833

PR #869, opened 2021-01-22
Generate metrics by service loading instead of hardcoded list

emmaholmbergohlsson:
All metrics now extends Metrics and can be service loaded,
to avoid having a hardcoded list of metrics classes that we forget to
update.
This will:
* add some new metric that weren't found before
* change the order the metrics are generated in
* change the names of some metrics that got format
specifiers included before


emmaholmbergohlsson:
Need to go in together with https://github.com/neo-technology/neo4j/pull/8714 and https://github.com/neo-technology/neo4j-manual-modeling/pull/1774

davidoliverSP2 (approved):
thanks @emmaholmbergohlsson 

I can merge this when the Neo4j PR is approved and merged.

PR #882, opened 2021-01-27
Autogenerate function descriptions

LinneaAndersson:
In this PR we have created a function to autogenerate the functions tables where we list all functions by category. (this page: https://neo4j.com/docs/cypher-manual/current/functions/). During the build the `GenerateFunctionDescriptions ` creates a document `functionDescription-descriptions-complete.adoc` that contains the tables. The tables are then copied to the file `cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc`.

There has been a few changes and I would like your input on whether or not these changes are okay. 

#### Before:
Screenshot 2021-02-03 at 09 57 38

#### After:
Screenshot 2021-02-03 at 09 58 05

#### Before
We created multiple rows in the list for the same function but with different input, as for an example the `date` function. Each row links to one example query. 
 
Screenshot 2021-02-03 at 11 25 47

 
#### After
One row per function - linking to the first example of the function.
Screenshot 2021-02-03 at 11 26 02

### 3) Functions are listed in order

In most cases the function tables are in the same order in this PR and in the current 4.3. But this is not the case in:

- LOAD CSV functions (`file()` and `linenumber()` )
- Temporal functions - duration (`inMonth()`, `inDays()`, `inSeconds()`), is this ok?

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/glossary.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/glossary.asciidoc
@@ -169,7 +169,7 @@ This section comprises a glossary of all the keywords -- grouped by category and
 |<<functions-haversin, haversin()>>             | Trigonometric     | Returns half the versine of a number.
 |<<functions-head, head()>>                      | Scalar            | Returns the first element in a list.
 |<<functions-id, id()>>                          | Scalar            | Returns the id of a relationship or node.
-|<<functions-isEmpty, isEmpty()>>                          | Predicate            | Returns true if the given list or map contains no elements or if the given string contains no characters.
+|<<functions-isempty, isEmpty()>>                          | Predicate            | Returns true if the given list or map contains no elements or if the given string contains no characters.
 |<<functions-keys, keys()>>                      | List              | Returns a list containing the string representations for all the property names of a node, relationship, or map.
 |<<functions-labels, labels()>>                  | List              | Returns a list containing the string representations for all the labels of a node.
 |<<functions-last, last()>>                      | Scalar            | Returns the last element in a list.
@@ -219,7 +219,7 @@ This section comprises a glossary of all the keywords -- grouped by category and
 |<<functions-reverse-list, reverse()>>           | List              | Returns a list in which the order of all elements in the original list have been reversed.
 |<<functions-reverse, reverse()>>                | String            | Returns a string in which the order of all characters in the original string have been reversed.
 |<<functions-right, right()>>                    | String            | Returns a string containing the specified number of rightmost characters of the original string.
-|<<functions-round1, round()>>                   | Numeric           | Returns the value of the given number rounded to the nearest integer, with half-way values always rounded up.
+|<<functions-round, round()>>                   | Numeric           | Returns the value of the given number rounded to the nearest integer, with half-way values always rounded up.
 |<<functions-round2, round(), with precision>>   | Numeric           | Returns the value of the given number rounded with the specified precision, with half-values always being rounded up.
 |<<functions-round3, round(), with precision and rounding mode>>  | Numeric           | Returns the value of the given number rounded with the specified precision and the specified rounding mode.
 |<<functions-rtrim, rTrim()>>                    | String            | Returns the original string with trailing whitespace removed.
--- a/cypher/cypher-docs/src/docs/dev/glossary.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/glossary.asciidoc
@@ -206,7 +206,7 @@ This section comprises a glossary of all the keywords -- grouped by category and
 |<<functions-pi, pi()>>                         | Trigonometric     | Returns the mathematical constant _pi_.
 |<<functions-point-cartesian-2d, point() - Cartesian 2D>>  | Spatial           | Returns a 2D point object, given two coordinate values in the Cartesian coordinate system.
 |<<functions-point-cartesian-3d,point() - Cartesian 3D>> | Spatial           | Returns a 3D point object, given three coordinate values in the Cartesian coordinate system.
-|<<functions-point-wgs84-2d, point() - WGS 84 2D>>          | Spatial           | Returns a 2D point object, given two coordinate values in the WGS 84 coordinate system.
+|<<functions-point, point() - WGS 84 2D>>          | Spatial           | Returns a 2D point object, given two coordinate values in the WGS 84 coordinate system.
 |<<functions-point-wgs84-3d,point() - WGS 84 3D>> | Spatial         |  Returns a 3D point object, given three coordinate values in the WGS 84 coordinate system.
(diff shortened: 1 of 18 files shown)


cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -206,7 +206,7 @@ This section comprises a glossary of all the keywords -- grouped by category and
 |<<functions-pi, pi()>>                         | Trigonometric     | Returns the mathematical constant _pi_.
 |<<functions-point-cartesian-2d, point() - Cartesian 2D>>  | Spatial           | Returns a 2D point object, given two coordinate values in the Cartesian coordinate system.
 |<<functions-point-cartesian-3d,point() - Cartesian 3D>> | Spatial           | Returns a 3D point object, given three coordinate values in the Cartesian coordinate system.
-|<<functions-point-wgs84-2d, point() - WGS 84 2D>>          | Spatial           | Returns a 2D point object, given two coordinate values in the WGS 84 coordinate system.
+|<<functions-point, point() - WGS 84 2D>>          | Spatial           | Returns a 2D point object, given two coordinate values in the WGS 84 coordinate system.
```
martin-neotech: I would use the `crs` to clean up the `point()` function links.

https://neo4j.com/docs/cypher-manual/current/functions/spatial/#functions-point-cartesian-2d for example have the `csr` `cartesian`

So to cleanup the point() function links would result in:

+ `functions-point-cartesian`
+ `functions-point-cartesian-3d`
+ `functions-point-wgs-84`
+ `functions-point-wgs-84-3d`

And probably do the point function links cleanup in a seperate PR after this one.

cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -28,74 +28,79 @@ Related information may be found in <<query-operators>>.
...
 [options="header"]
 |===
-|Function                         |   Description
-|<<functions-all,all()>>          | Tests whether the predicate holds for all elements in a list.
-|<<functions-any,any()>>          | Tests whether the predicate holds for at least one element in a list.
-|<<functions-exists,exists()>>    | Returns true if a match for the pattern exists in the graph, or if the specified property exists in the node, relationship or map.
-|<<functions-isEmpty, isEmpty()>> | Returns true if the given list or map contains no elements or if the given string contains no characters.
-|<<functions-none,none()>>        | Returns true if the predicate holds for no element in a list.
-|<<functions-single,single()>>    | Returns true if the predicate holds for exactly one of the elements in a list.
+| Function | Signature | Description
```
martin-neotech, line 38: Is it possible to use this asciidoc layout style instead?


```
[options="header"]
|===
| Function | Signature | Description

| <<functions-all,`all()`>>
| `all(variable :: VARIABLE IN list :: LIST OF ANY? WHERE predicate :: ANY?) :: (BOOLEAN?)`
| Returns true if the predicate holds for all elements in the given list.

1.3+| <<functions-isempty,`isEmpty()`>>
| `isEmpty(input :: LIST? OF ANY?) :: (BOOLEAN?)`
| Checks whether a list is empty.

//isEmpty()
| `isEmpty(input :: MAP?) :: (BOOLEAN?)`
| Checks whether a map is empty.

//isEmtpy()
| `isEmpty(input :: STRING?) :: (BOOLEAN?)`
| Checks whether a string is empty.

|===
```

martin-neotech (commented):
This is great! Only some minor comments.

functions/src/main/java/org/neo4j/doc/FunctionDescriptionsGenerator.java
```diff
@@ -0,0 +1,284 @@
...
+            return format( "| %s | %s%n", signature, description );
+        }
+
+        public String firstRow( int sameNameCount )
+        {
+            String referenceName = "functions-" + name.toLowerCase().replace( "()", "" ).replace( '.', '-' );
+            if (referenceExceptions.containsKey( referenceName )) {
+                referenceName = referenceExceptions.get( referenceName );
+            }
+            return format( "1.%s+| <<%s,%s>>  | %s | %s%n",
```
martin-neotech: To make it code styled

```
format( "1.%s+| <<%s,`%s`>>  | `%s` | %s%n",
```

functions/src/main/java/org/neo4j/doc/FunctionDescriptionsGenerator.java
```diff
@@ -0,0 +1,254 @@
...
+        public int hashCode()
+        {
+            int result = name.hashCode();
+            result = 31 * result + signature.hashCode();
+            return result;
+        }
+
+        public String row()
+        {
+            return "| <<functions-" + name.toLowerCase().replace( "()", "" ) + "," + name + ">>  | " + description + "\n";
```
pontusmelke: I am not sure if this is a problem but if running on windows the `\n` might cause issues, maybe use `System.lineBreak()` instead.

functions/src/main/java/org/neo4j/doc/FunctionDescriptionsGenerator.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.Map;
+
+import org.neo4j.dbms.api.DatabaseManagementService;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Result;
+import org.neo4j.graphdb.Transaction;
+
+import static org.neo4j.configuration.GraphDatabaseSettings.DEFAULT_DATABASE_NAME;
+
+public class FunctionDescriptionsGenerator
```
pontusmelke, line 42: This file uses `\n` which might cause rendering issues on windows. Maybe use `System.lineBreak()` or `String.format("...%n")`

pontusmelke (approved):
Looks good to me. 

My asciidoc is a bit rusty so I didn't look too closely at those changes.

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
LinneaAndersson: Do you mean that you would like to split the `point` function into 4 different rows in the functions table, one for each crs? (like it was before)

martin-neotech: Yes please.

LinneaAndersson: I reverted the point functions in the table. The problems I can see with this is that: 
1) We need to do this manually everytime we update the function tables. 
2) It is maybe a bit weird to have the same function (with same signature) on multiple rows in the table.

What do you think? I can leave it like this if you think it's okay :)

martin-neotech (approved):
Nice!

PR #888, opened 2021-01-28
Updated read-only example

hugofirth:
Single line change, needed as the internal `read_only` configuration value is no longer encouraged. In particular, `dbms.read_only` doesn't work with clusters, whilst embedded now **does**. `dbms.databases.default_to_read_only` and friends are the preferred replacements.


AlexicaWright (approved):
This looks good to me and I can build it on my machine, but the tests are still failing on TC.

davidoliverSP2:
@hugofirth please could you rebase this PR? tests _should_ be green with latest updates

PR #889, opened 2021-02-01
prettifier disabled

martin-neotech:
Also fixed some string syntax and cypher queries.


recrwplay (approved):
Spot checked codeblocks that are displayed differently as a result of this PR and they are improved now.

PR #891, opened 2021-02-01
4.0 licence text update

martin-neotech:
Removed year in licence and added licence text for those scala files that did not have one.


recrwplay (approved):
This looks fine to me as long as we don't need to specify any years in the copyright statement.

martin-neotech:
> This looks fine to me as long as we don't need to specify any years in the copyright statement.

I asked Chris Gioran how it should look without the years, so it is confirmed by him.

PR #933, opened 2021-02-18
4.0 cypher query syntax fix part 2

martin-neotech:
fixed indentation etc.


martin-neotech (commented):
Cypher format looks good.

PR #963, opened 2021-02-25
SHOW USER/ROLE PRIVILEGES allows yield and where

Hunterness:
The syntax was correct in the privilege section but missing in the user and role sections.

This is already fixed in later versions but was apparently missed here.


Lojjs (approved):
Wasn't able to build it due to an unrelated test failure, but looking at the PR this looks good.

martin-neotech (approved):

PR #964, opened 2021-02-25
`SHOW ... WHERE ... RETURN ...` without the `YIELD` is not allowed

Hunterness:
Clarify this in the documentation.


Lojjs (approved):
I got an unrelated error when trying to build it, but just looking at the PR I think it looks correct now

martin-neotech (approved):

PR #1007, opened 2021-03-08
removed link for timestamp

martin-neotech:
the foot note cant render properly with a xref link inside.


AlexicaWright (approved):
Looks much better! Thank you @martin-neotech 👍

PR #1045, opened 2021-03-19
fixed xref comma text issue

martin-neotech:


recrwplay (approved):
I think this information could be presented more cleanly (there are really only three different instances, and the link target in any column is the same for any row in that column. However, that is outside the scope of this fix, which outputs the link text correctly.

PR #1068, opened 2021-03-31
Add `FULLTEXT` filtering to `SHOW INDEXES`

Hunterness:
~~Currently have broken tests due to relationship property indexes now being blocked (https://github.com/neo-technology/neo4j/pull/9567), fixed by https://github.com/neo4j/neo4j-documentation/pull/1069.~~

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
@@ -186,8 +186,11 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
       title = "Listing indexes with filtering",
       text =
         """
-          |One way of filtering the output from `SHOW INDEXES` is to use the `BTREE` keyword to only list `BTREE` indexes.
-          |Another is to use the `WHERE` clause, for example to only show indexes not belonging to constraints.""".stripMargin,
+          |One way of filtering the output from `SHOW INDEXES` by index type is the use of type keywords,
+          |listed in the <<administration-indexes-syntax, syntax table>>.
+          |As an example, to show only B-tree indexes, use `SHOW BTREE INDEXES`.
+          |Another, more flexible way of filtering the output is to use the `WHERE` clause.
+          |An example is to only show indexes not belonging to constraints.""".stripMargin,
       prepare = _ => executePreparationQueries(List("create index for ()-[r:KNOWS]-() on (r.since)")),
       queryText = "SHOW BTREE INDEXES WHERE uniqueness = 'NONUNIQUE'",
       optionalResultExplanation =
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -54,6 +54,7 @@ Replacement syntax for deprecated and removed features are also indicated.
 | `ON DEFAULT DATABASE` for privilege commands  | Syntax        | Deprecated    | Replaced by `ON HOME DATABASE`
 | `ON DEFAULT GRAPH` for privilege commands     | Syntax        | Deprecated    | Replaced by `ON HOME GRAPH`
 | `CREATE INDEX [name] FOR ()-[r:TYPE]-() ON (r.prop)` | Functionality | Added | Allows creating indexes on relationships.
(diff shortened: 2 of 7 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -186,8 +186,11 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
       title = "Listing indexes with filtering",
       text =
         """
-          |One way of filtering the output from `SHOW INDEXES` is to use the `BTREE` keyword to only list `BTREE` indexes.
-          |Another is to use the `WHERE` clause, for example to only show indexes not belonging to constraints.""".stripMargin,
+          |One way of filtering the output from `SHOW INDEXES` by index type is the type keywords,
```
SaschaPeukert:
```suggestion
          |One way of filtering the output from `SHOW INDEXES` by index type is the use of type keywords,
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
```diff
@@ -186,8 +186,11 @@ class SchemaIndexTest extends DocumentingTestBase with QueryStatisticsTestSuppor
...
+          |listed in the <<administration-indexes-syntax, syntax table>>.
+          |As an example, to show only B-tree indexes, use `SHOW BTREE INDEXES`.
+          |Another, more flexible way of filtering the output is to use the `WHERE` clause.
+          |For example to only show indexes not belonging to constraints.""".stripMargin,
```
SaschaPeukert: This sentence sounds like its not finished. Like its leading to an example that does not follow

SaschaPeukert (changes requested):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
Hunterness: should update that for constraints as well then, think I just took the sentence from there

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
Hunterness: There is supposed to be an example after? (the query below?)

SaschaPeukert: Thats my point. The query below does not fit the text "only show indexes not belonging to constraints", right?

Hunterness: - ```As an example, to show only B-tree indexes, use `SHOW BTREE INDEXES`.```
- `For example to only show indexes not belonging to constraints.`

`SHOW BTREE INDEXES WHERE uniqueness = 'NONUNIQUE'` shows all BTREE indexes not belonging to any constraints

Hunterness: the example combines both filtering versions

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
Hunterness: added to the constraints PR

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SchemaIndexTest.scala
SaschaPeukert: I am legit okay with any of the two as they both don't seem to reference the next query :-)

Hunterness: not sure what you mean?

the query in the example is `SHOW BTREE INDEXES WHERE uniqueness = 'NONUNIQUE'` with
the WHERE clause is filtering on indexes without constraints and the BTREE keyword filters on type which is the 2 things the text describes?

SaschaPeukert: Ah okay, slight miscommunication, now I understand :-D So my problem is that the last sentence does not sound like its belonging to the rest of the paragraph but instead sounds like it's something new. And if you read it like that, then the query underneath does not fit.

SaschaPeukert (approved):

Hunterness:
Will be included in https://github.com/neo4j/neo4j-documentation/pull/1075

PR #1069, opened 2021-03-31
Enable relationship property indexes feature flag for docs

emmaholmbergohlsson:
A new feature flag was introduced to hide the relationship property index feature until it is complete.
This enables the flag in tests and doc creation so that the documentation of those indexes that had
already been added still work.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-search-performance/index.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-search-performance/index.asciidoc
@@ -219,10 +219,6 @@ include::../indexes/create-a-single-property-index-for-nodes.asciidoc[leveloffse
 
 include::../indexes/create-a-single-property-index-for-relationships.asciidoc[leveloffset=+1]
 
-[NOTE]
-Relationship property indexes is a feature under development. If you want to try it out you need to set `unsupported.dbms.enable_relationship_property_indexes` setting in the _neo4j.conf_ file.
-Please note that no guarantees are made about functionality until the feature is complete.
-
 include::../indexes/create-a-single-property-index-only-if-it-does-not-already-exist.asciidoc[leveloffset=+1]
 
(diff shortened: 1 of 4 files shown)


emmaholmbergohlsson:
Discussed it with @Hunterness and we decided that we wanted to keep the documentation of the new indexes even though the feature isn't complete yet. The feature will be done before 4.3 is released and then we remove the NOTEs again

Hunterness (approved):

cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-search-performance/index.asciidoc
```diff
@@ -219,6 +219,10 @@ include::../indexes/create-a-single-property-index-for-nodes.asciidoc[leveloffse
 
 include::../indexes/create-a-single-property-index-for-relationships.asciidoc[leveloffset=+1]
 
+[NOTE]
+Relationship property indexes is a feature under development. If you want to try it out you need to set `unsupported.dbms.enable_relationship_property_indexes` setting in the _neo4j.conf_ file.
+Please note that no guarantees are made about functionality until the feature is complete.
+
```
davidoliverSP2: Currently reviewing this message with PM team. 

With this being "preview" docs prior to release, I don't think there is any necessity to declare this as work in progress. If this feature is only surfaced in Aura, then we should add this comment to the internal, unsupported documentation here: https://sites.google.com/neotechnology.com/dbms-internal-documentation/internal-configuration?authuser=1

emmaholmbergohlsson: Okay, we don't actually want anybody trying it out yet. We just didn't want to remove the documentation just to have to put it back again later, so we thought a note might be necessary

davidoliverSP2: Perhaps it might be best to add this in as comments then, and add something to the internal docs?

Hunterness (approved):

PR #1077, opened 2021-04-07
Adapt ReadOnlyDocTest

MishaDemianenko:


davidoliverSP2 (approved):
Thanks @MishaDemianenko

This builds fine, and looks good , but I can't find where this surfaces in the docs. Do you know if this will update a page in a manual?

MishaDemianenko:
Hi @davidoliverSP2 so based on tags that are used in the test it looks like this page is using this test: https://github.com/neo-technology/neo4j-manual-modeling/blob/4.3/javaReference/asciidoc/embedded/setup.adoc

PR #1138, opened 2021-06-08
removed old file

martin-neotech:
cypher/cypher-docs/src/docs/dev/ql/merge/index.asciidoc is not used.
query-merge is already in MergeTest.scala


recrwplay (approved):
Verified that the content is a duplicate of MergeTest.scala

PR #1143, opened 2021-06-11
fixed leveloffset

martin-neotech:


recrwplay (approved):
Thanks for resolving these!

Verified by local build.

PR #1149, opened 2021-06-17
Cleanup after Prettifier was ditched.

sherfert:
#889 removed the use of the Prettifier for formatting queries.
It missed cleaning up the code around that, though.


davidoliverSP2 (approved):
Thanks @sherfert 

Tests all green after a rebase.

PR #1150, opened 2021-06-17
Generate internal settings documentation

Roxling:
Will output an additional `/docs/ops/internal-settings.adoc` with (only) the settings marked as internal

### DIFF
--- a/config-docs/pom.xml
+++ b/config-docs/pom.xml
@@ -107,6 +107,18 @@
                                     <arg file="${project.build.directory}/docs/ops/all-settings.adoc" />
                                 </java>
 
+                                <!-- All internal settings -->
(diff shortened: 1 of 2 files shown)


martin-neotech (approved):
Thanks

davidoliverSP2 (approved):
Thanks @Roxling 

We can add some instructions here that this output needs to be converted to html (or a readable format) and added to the internal documentation site for each drop/release.

Do you know if there are any blockers to cherry picking this back to 4.3?

Roxling:
@davidoliverSP2 This should work just as fine on 4.3.

 Generating an additional output format can probably pretty easily be added to this tool and be selected per output file in the pom, or just use any available converter tool.

PR #1153, opened 2021-07-01
Improved messaging around duration components

MarkWoulfeNeo4j:
Merged two tables into one and removed duplicate and confusing conversion list.

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TemporalTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TemporalTest.scala
@@ -473,51 +473,31 @@ class TemporalTest extends DocumentingTest {
       }
       section("Accessing components of durations", "cypher-temporal-accessing-components-durations") {
         p(
-          """A _Duration_ can have several components.
-            |These are categorized into the following groups:
+          """A _Duration_ can have several components, each categorized into _Months_, _Days_, and _Seconds_ groups.
             |
-            |[options="header"]
+            |Components of _Duration_ values are truncated within their component groups as follows:
+            |[options="header", cols="2,3,2,1,3"]
             ||===
-            || Component group | Constituent components
-            || Months | _Years_, _Quarters_ and _Months_
-            || Days | _Weeks_ and _Days_
-            || Seconds | _Hours_, _Minutes_, _Seconds_, _Milliseconds_, _Microseconds_ and _Nanoseconds_
+            || Component Group | Component | Description | Type | Details
+            |.3+| _Months_| `duration.years` | The total number of _years_ | Integer | Each set of `4` _quarters_ is counted as `1` _year_; each set of `12` _months_ is counted as `1` _year_.
+            || `duration.quarters` | The total number of _quarters_ | Integer | Each _year_ is counted as `4` _quarters_; each set of `3` _months_ is counted as `1` _quarter_.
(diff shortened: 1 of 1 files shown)


MarkWoulfeNeo4j:
Here is an image of what the changes look like to save you having to go through the build process

image

davidoliverSP2 (approved):
Looks good, thanks @MarkWoulfeNeo4j

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TemporalTest.scala
```diff
@@ -473,51 +473,31 @@ class TemporalTest extends DocumentingTest {
...
+            || `duration.months` | The total number of _months_ | Integer | Each _year_ is counted as `12` _months_; each_quarter_ is counted as `3` _months_.
+            |.2+| _Days_ | `duration.weeks` | The total number of _weeks_ | Integer | Each set of `7` _days_ is counted as `1` _week_.
+            || `duration.days` | The total number of _days_ | Integer | Each _week_ is counted as `7` _days_.
+            |.6+| _Seconds_ | `duration.hours` | The total number of _hours_ | Integer | Each set of `60` _minutes_ is counted as `1` _hour_; each set of `3600` _seconds_ is counted as `1` _hour_.
+            || `duration.minutes` | The total number of _minutes_ | Integer | Each _hour_ is counted as `60` _minutes_; each set of `60` _seconds_ is counted as `1` _minute_.
+            || `duration.seconds` | The total number of _seconds_ | Integer | Each _hour_ is counted as `3600` _seconds_; each _minute_ is counted as `60` _seconds_.
+            || `duration.milliseconds` | The total number of _milliseconds_ | Integer | Each set of `1000` _milliseconds_ is counted as `1` _second_.
```
Lojjs, line 490: This is a bit tricky. The reason we did not include any info on the last rows was a simplification. If this column should be complete, it should contain info on all the other components in the same group, since all those can be converted to each other. So the ones for day and above are complete, but all the time based ones would need to change, e.g.

hour: Each set of `60` _minutes_ is counted as `1` _hour_; each set of `3600` _seconds_ is counted as `1` _hour_; each set of `3 600 000` milliseconds is counted as `1` _hour_; each set of `3 600 000 000` microseconds is counted as `1` _hour_; each set of `3 600 000 000 000` nanoseconds is counted as `1` _hour_.

millisecond: Each _hour_ is counted as `3 600 000` _milliseconds_; each _minute_ is counted as `60000` _milliseconds_; each _second_ is counted as `1000` _milliseconds_; each set of `1000` _microseconds_ is counted as `1` _millisecond_; each set of `1 000 000` _nanoseconds_ is counted as `1` _millisecond_.

But this get very long and probably hard to read in a table. Don't know if you have any other better suggestion on how to present this part?

Lojjs (changes requested):
I'm a bit unsure about how to best handle the Details columns, see separate comment. Other than that, the merging of the tables looks good.

JMHReif (approved):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TemporalTest.scala
MarkWoulfeNeo4j: Hey Louise,

Had a quick chat with @davidoliverSP2 and we couldn't come up with an elegant solution to this off the top of our heads. However, we believe that this is still an improvement over the existing material and will continue ahead with merging this PR. 

We can revisit this if any customers complain about a lack of completeness and any resulting confusion.

PR #1158, opened 2021-07-05
Described unit-subqueries.

arnefischereit:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -49,13 +50,19 @@ class CallSubqueryTest extends DocumentingTest {
...
+
+      p(
+        """Subqueries which end in a `RETURN` statement are called _returning subqueries_ while subqueries without such a return statement are called _unit
+          |subqueries_.""".stripMargin)
+      p("""A subquery is evaluated for each incoming input row. Every output row of a *returning subquery* is combined with the input row to build the result of
+          #the subquery.
+          #That means that a returning subquery will influence the number of rows.
           #If the subquery does not return any rows, there will be no rows available after the subquery.""".stripMargin('#'))
+      p(
+        """*Unit subqueries* on the other hand are thought to be called for their side-effects and not for their results and therefore do not interfere with
```
tobias-johansson: I think we can simplify this to just
```suggestion
        """*Unit subqueries* on the other hand are called for their side-effects and not for their results, and do therefore not influence
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -49,13 +50,19 @@ class CallSubqueryTest extends DocumentingTest {
...
+          #the results of the enclosing query.""".stripMargin('#'))
       p("There are restrictions on what queries are allowed as subqueries and how they interact with the enclosing query:")
```
tobias-johansson: Perhaps we can simplify this now
```suggestion
      p("There are restrictions on how subqueries interact with the enclosing query:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
 
-    section("Aggregation and side-effects", "subquery-aggregation") {
-      p("""Subqueries can be useful to do aggregations for each row and to isolate side-effects.
-          #This example query creates five `Clone` nodes for each existing person.
-          #The aggregation ensures that cardinality is not changed by the subquery.
-          #Without this, the result would be five times as many rows.""".stripMargin('#'))
+    section("Cardinality and aggregation", "subquery-aggregation") {
+      p(
+        """Returning subqueries change the cardinality of the query: The results of each subquery call are collected and together they give the result of the
+          |enclosing query.""".stripMargin)
```
tobias-johansson: Shouldn't we rather say that they are the result of the `CALL` clause
```suggestion
          |`CALL` clause.""".stripMargin)
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
+      p("""The following example lists each connected person that each person has:""")
```
tobias-johansson:
```suggestion
      p("""The following example finds the names of each person's friends:""")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
-          Map("p.name" -> "Alice", "numberOfClones" -> 5),
-          Map("p.name" -> "Bob", "numberOfClones" -> 5),
-          Map("p.name" -> "Charlie", "numberOfClones" -> 5),
-          Map("p.name" -> "Dora", "numberOfClones" -> 5)
+          Map("p.name" -> "Alice", "friend" -> "Bob"),
+          Map("p.name" -> "Bob", "friend" -> "Alice"),
+        )))) { resultTable() }
+      p(
+        """The number of results of the subquery changed the number of results of the enclosing query: Instead of 4 rows, one for each node), there are
+          |now 2 rows which were found for Alice and Bob respectively. The nodes for Charlie and Dora are not returned.""".stripMargin)
```
tobias-johansson: Let's be explicit, even though it is a sad truth :cry: 
```suggestion
          |now 2 rows which were found for Alice and Bob respectively. No rows are returned for Charlie and Dora since they have no friends.""".stripMargin)
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
+      p(
+        """Besides changing the set of matched nodes, which we can also express using a simple query, we can also use the subquery to aggregate on each of
+          |the matches of the enclosing query in isolation. As we get one result for each call of the subquery, the cardinality does not differ from the
+          |enclosing query's cardinality:""".stripMargin)
```
tobias-johansson: I'd like to suggest:
``We can also use subqueries to perform isolated aggregations. In this example we count the number of relationships each person has. As we get one result for each call of the subquery, the number of rows is the same, before and after the `CALL` clause:``

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
+        ResultAssertions(r => r.toSet should equal(Set(
+          Map("p.name" -> "Alice", "numberOfConnections" -> 2),
+          Map("p.name" -> "Bob", "numberOfConnections" -> 1),
+          Map("p.name" -> "Charlie", "numberOfConnections" -> 1),
+          Map("p.name" -> "Dora", "numberOfConnections" -> 0)
+        )))) { resultTable() }
+    }
+
+    section("Unit subqueries and side-effects", "subquery-unit") {
+      p("""Unit subqueries do not return any value and are therefore used for their side effects.""")
```
tobias-johansson: Perhaps we should use "rows" as the thing that is returned
```suggestion
      p("""Unit subqueries do not return any rows and are therefore used for their side effects.""")
```

tobias-johansson (changes requested):
I think we can simplify and clarify in a few places :smiley:

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
arnefischereit: Agreed. I would then also suggest to replace call with "invocation" in the line before to not use the word "call" twice with two different meanings.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
arnefischereit: I generally like your version better. Two comments:

- Should we write "As we get one _row_" as you have suggested elsewhere to talk about rows? And possibly we should also write "invocation" of the subquery? (see other comment)
- We use  the word "cardinality" in the title and elsewhere, so I thought I could also use it here to unify the wording.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
arnefischereit: I can change this. What I wanted to express was that we return rows of both the person and their friend respectively.

tobias-johansson: Right. How about
```suggestion
      p("""The following example finds the name of each person and the names of their friends:""")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
tobias-johansson: `s/call/invocation/` :+1: 
Just did a quick search of the docs. The word _cardinality_ does not appear anywhere in the cypher manual, but the word _rows_ does multiple times :shrug:

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
arnefischereit: Let's take that.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
arnefischereit: Right. I just realized that it was my change here, which introduced "cardinality" in the headline :see_no_evil: 

I now changed the headline to just read "Aggregations" instead of "Cardinality and aggregations".

tobias-johansson (changes requested):
I like it! Except now we sometimes say "evaluation" and sometimes "invocation" :smile:

arnefischereit:
> I like it! Except now we sometimes say "evaluation" and sometimes "invocation" smile

To me it feels like these two terms describe the two sides of the "subquery interface": From the outside, the subquery is invoked. On the inside, the statements that make up the subquery are evaluated.

If you think that it improves the readability, we can align this wording. I personally would prefer "invocation" but I am open to your suggestion.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
 
-    section("Aggregation and side-effects", "subquery-aggregation") {
-      p("""Subqueries can be useful to do aggregations for each row and to isolate side-effects.
-          #This example query creates five `Clone` nodes for each existing person.
-          #The aggregation ensures that cardinality is not changed by the subquery.
-          #Without this, the result would be five times as many rows.""".stripMargin('#'))
+    section("Aggregations", "subquery-aggregation") {
+      p(
+        """Returning subqueries change the number of results of the query: The results of each subquery invocation are collected and together they give the
+          |result of the `CALL` clause.""".stripMargin)
```
tobias-johansson: ``Returning subqueries change the number of results of the query: The result of the `CALL` clause is the combined result of evaluating the subquery for each input row.``

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
-          Map("p.name" -> "Alice", "numberOfClones" -> 5),
-          Map("p.name" -> "Bob", "numberOfClones" -> 5),
-          Map("p.name" -> "Charlie", "numberOfClones" -> 5),
-          Map("p.name" -> "Dora", "numberOfClones" -> 5)
+          Map("p.name" -> "Alice", "friend" -> "Bob"),
+          Map("p.name" -> "Bob", "friend" -> "Alice"),
+        )))) { resultTable() }
+      p(
+        """The number of results of the subquery changed the number of results of the enclosing query: Instead of 4 rows, one for each node), there are
+          |now 2 rows which were found for Alice and Bob respectively. No rows are returned for Charlie and Dora since they have no friends in our example
```
tobias-johansson, line 156: :+1: :smile:

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CallSubqueryTest.scala
```diff
@@ -126,23 +135,59 @@ class CallSubqueryTest extends DocumentingTest {
...
+          |graph.""".stripMargin)
+      p(
+        """We can also use subqueries to perform isolated aggregations. In this example we count the number of relationships each person has. As we get one row
+          | for each invocation of the subquery, the number of rows is the same, before and after the `CALL` clause:""".stripMargin)
```
tobias-johansson:
```suggestion
          | from each evaluation of the subquery, the number of rows is the same, before and after the `CALL` clause:""".stripMargin)
```

tobias-johansson (commented):
Adding suggestions for what it _would_ look like if we changed all `invocation` to `evaluation`. Let's discuss on slack :)

arnefischereit:
[Document Unit Subquery CALLs](https://trello.com/c/OF5PUoCh/527-document-unit-subquery-calls)

tobias-johansson (approved):

PR #1180, opened 2021-07-14
Add a link so examples how to construct timezones can be found easier.

sherfert:


arnefischereit (approved):
This change is fine.

However, we originally wanted to document why something like `GMT+03:00` was accepted while it was not part of the formats for specifying time zones.

We can take the discussion elsewhere if needed.

AlexicaWright (approved):
👍  LGTM and link works fine.

PR #1184, opened 2021-07-15
Minimum length of var length patterns when length omitted

arnefischereit:


arnefischereit:
[Document * in paths](https://trello.com/c/r5NqGKHL/555-document-in-paths)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PatternTest.scala
```diff
@@ -199,7 +200,7 @@ class PatternTest extends DocumentingTest {
           |(a)-[*2]->(b)
           |----
           |""")
-      p("""This describes a graph of three nodes and two relationship, all in one path (a path of length 2).
+      p("""This describes a subgraph of three nodes and two relationships, all in one path (a path of length 2).
```
tobias-johansson: A few lines down also has `graph` https://github.com/neo4j/neo4j-documentation/blob/bd5b4940e295e4aaa42919b0b6be862ed413e243/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PatternTest.scala#L218

Should they both be `subgraph` or both be `graph`?
One view is that a match pattern describes a `graph`, and then we search for `subgraphs` that are "similar" in a larger graph.

tobias-johansson (changes requested):
Looks good!
I just want to discuss some terminology :)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PatternTest.scala
arnefischereit: I think that is the better way to look at it. I will revert that change.

tobias-johansson (approved):
:+1:

renetapopova (approved):
Looks good and builds fine, thanks @arnefischereit!

PR #1185, opened 2021-07-15
Allocate web containers in temporary directories instead of system directory

MishaDemianenko:


davidoliverSP2 (approved):
Thanks @MishaDemianenko - I built this locally and was successful. I'll rerun the PR build and await a green test before merging. 

This resolves our local build issues on the `dev` branch. I assume we can cherry pick this back to 4.3? 

Our local builds of 4.2 and earlier appear to be stable - is there a reason for this problem only affecting 4.3 and 4.4 builds?

MishaDemianenko:
@davidoliverSP2 maybe it was updated to be like that in those versions or some other dependency changed. Absolutely we should be able to cherry pick this for older versions.

PR #1196, opened 2021-07-21
Remove example block delimeter from end of file

recrwplay:
A warning is reported in the HTML and PDF builds.

The HTML build is unaffected because it is the end of the file.

The PDF build is affected because the PDF is built as if it were a single asciidoc file, so all the remaining content after this example block delimiter is treated as a giant example block, which in turn means the table of contents ends at this point.


davidoliverSP2 (approved):
good spot

PR #1197, opened 2021-07-21
Make check marks uniform

davidoliverSP2:


davidoliverSP2:
Hey @recrwplay - could you review this PR?

For the scala tests, it wasn't clear how I could inject the `:check-mark: icon:check[]` required to point at the icon, so I just used the character ✔ . This isn't very consistent, but didn't think it warranted a lot of investigation, especially as we'd like more control over these pages in the near future. If you know of a quick way to work around that though, please let me know!

recrwplay (approved):
LGTM

PR #1202, opened 2021-07-22
Move and rename the Database management section

davidoliverSP2:
Depends on https://github.com/neo-technology/neo4j-manual-modeling/pull/2283

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/administration.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/administration.asciidoc
@@ -6,6 +6,12 @@
 This chapter explains how to use Cypher to administer Neo4j databases, for the purpose of managing indexes, and managing security.
 --
 
+Neo4j supports the management of multiple databases within the same DBMS.
+The metadata for these databases, including the associated security model, is maintained in a special database called the `system` database.
+Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system.
+This includes all commands related to defining and managing the security model: users, roles and privileges.
+The administrative commands that are specific to the schema of an individual database are still executed against that specific database.
+
 * <<administration-indexes-search-performance, Indexes for search performance>>
 ** <<administration-indexes-introduction, Introduction>>
 ** <<administration-indexes-syntax, Syntax>>
--- a/cypher/cypher-docs/src/docs/dev/administration.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/administration.asciidoc
@@ -3,24 +3,8 @@
 
 [abstract]
 --
-This chapter explains how to use Cypher to administer Neo4j databases, such as creating databases, managing indexes and constraints, and managing security.
+This chapter explains how to use Cypher to administer Neo4j database, for the purpose of managing indexes, and managing security.
 --
-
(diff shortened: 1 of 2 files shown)


davidoliverSP2:
[Move the Database Management section out of the Cypher Administration section](https://trello.com/c/1Mr5ZxXV/3573-move-the-database-management-section-out-of-the-cypher-administration-section)

martin-neotech (approved):

Hunterness (commented):
You should probably update `cypher/cypher-docs/src/docs/dev/administration.asciidoc` as well (see the updates from https://github.com/neo4j/neo4j-documentation/pull/1159/files#diff-1929586aacfce12b0ddd2dedb236409ee5e6f4c951dd3dc09b0600fd1fd1d100)

cypher/cypher-docs/src/docs/dev/administration.asciidoc
```diff
@@ -3,25 +3,9 @@
...
 --
-This chapter explains how to use Cypher to administer Neo4j databases, such as creating databases, managing indexes and constraints, and managing security.
+This chapter explains how to use Cypher to administer Neo4j databases, for the purpose of managing indexes, and managing security.
 --
 
-Neo4j supports the management of multiple databases within the same DBMS.
-The metadata for these databases, including the associated security model, is maintained in a special database called the `system` database.
-Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system.
-This includes all commands related to managing multiple databases as well as all commands for defining the security model: users, roles and privileges.
-The administrative commands that are specific to the schema of an individual database are still executed against that specific database.
```
Hunterness: A lot of this is still valid... really only the part about managing databases are not.

```
Neo4j supports the management of multiple databases within the same DBMS.
The metadata for these databases, including the associated security model, is maintained in a special database called the `system` database.
Most administrative commands must be executed against the `system` database because they involve editing the metadata for the entire system.
This includes all commands related to defining  and managing the security model: users, roles and privileges.
The administrative commands that are specific to the schema of an individual database are still executed against that specific database.
```

However it should probably be stated in the new database chapter that the commands should be run against the system database...

davidoliverSP2: yep, I thought it was covered reasonably in the introduction to the Databases chapter:
Screenshot 2021-07-23 at 09 25 35

What do you think? I could incorporate some of the wording in this removed section?

As the Administration chapter itself is due to be removed, I thought it wasn't worth replacing this text with anything specific to indexes or security.

Hunterness: I think the databases are covered then yes.

I only rewrote 
`This includes all commands related to managing multiple databases as well as all commands for defining the security model: users, roles and privileges.` 
to 
`This includes all commands related to defining  and managing the security model: users, roles and privileges.`
the rest of the intro for admin section remained the same

davidoliverSP2: ah cool - I'll return that rewritten intro then (at least until we move the indexes & security!)

Sorry, I had missed that revision in your comment above!

Hunterness (approved):

PR #1203, opened 2021-07-22
4.4 cleanup temporary test folders and files after building

martin-neotech:
This PR fixes the issue that there are files generated by the build that is of significant size ~ 5 GB in the /tmp directory.

There is a cleanup hook that removes all folders that are created as temporary.

Removed unused test that was generating empty folders in the /tmp directory.


davidoliverSP2 (approved):
Local builds are good 👍

PR #1228, opened 2021-08-06
Generate internal settings documentation

davidoliverSP2:
FYI @recrwplay


recrwplay:
For some reason this build fails with 'not enough space' _every_ time I try to run it on TeamCity, but I was able to run it locally.

martin-neotech:
It is green now. Do not forget to add instructions how this should be updated for a new version.

recrwplay:
The output file looks good to me

PR #1232, opened 2021-08-24
Update so limited result table don't need columns

Hunterness:
since column order has been fixed it can use the columns from the result instead of needing to list all columns explicitly

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/tooling/DocBuilder.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/tooling/DocBuilder.scala
@@ -130,19 +130,20 @@ trait DocBuilder {
     queryScope.addContent(new TablePlaceHolder(queryScope.assertions, queryScope.params:_*))
   }
 
-  /**Print a smaller version of the result without changing the query.
+  /** Print a smaller version of the result without changing the query.
    * Can limit the amount of printed rows and columns.
    * Will print the given columns in the given order.
    *
-   * @param wantedColumns the columns to be displayed, in the wanted display order
-   * @param rows          the number of result rows to display
+   * @param maybeWantedColumns the columns to be displayed, in the wanted display order.
+   *                           If no column list is given, all columns will be displayed (this is default).
+   * @param rows               the number of result rows to display
    */
-  def limitedResultTable(wantedColumns: List[String], rows: Int = 10): Unit = {
+  def limitedResultTable(rows: Int = 10, maybeWantedColumns: Option[List[String]] = None): Unit = {
     assert(rows > 0, "Cannot display less than one row")
     val queryScope = scope.collectFirst {
       case q: QueryScope => q
     }.get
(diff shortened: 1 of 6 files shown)


Lojjs (approved):
Looks fine to me. Tried to build it and the output looks like before, which I guess was the intention :)

Hunterness:
> Looks fine to me. Tried to build it and the output looks like before, which I guess was the intention :)

yes, that is intended :)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/tooling/DocBuilder.scala
```diff
@@ -130,19 +130,20 @@ trait DocBuilder {
...
    * Will print the given columns in the given order.
    *
-   * @param wantedColumns the columns to be displayed, in the wanted display order
-   * @param rows          the number of result rows to display
+   * @param maybeWantedColumns the columns to be displayed, in the wanted display order.
+   *                           If no column list is given, all columns will be displayed (this is default).
+   * @param rows               the number of result rows to display
    */
-  def limitedResultTable(wantedColumns: List[String], rows: Int = 10): Unit = {
+  def limitedResultTable(rows: Int = 10, maybeWantedColumns: Option[List[String]] = None): Unit = {
```
martin-neotech, line 141: Why do we even have this?

It is possible to add a `LIMIT` with Cypher.

Now it is possible to have an example and show an "incorrect" result.

Is there an example in the Cypher Manual where the output is actually limited?

 How does the user then know that the result has been limited?
 
 e.g:
 
 ```
"x" | "y"
1   | "a"
2   | "b"
... example output has been limited.
```

martin-neotech (approved):
Looks good, thanks.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/tooling/DocBuilder.scala
martin-neotech: Tested the limit on `SHOW FUNCTIONS`.

https://neo4j.com/docs/cypher-manual/4.3/clauses/listing-functions/#_listing_all_functions

It will basically obfuscate the actual result.

So if the real result returned is 20 rows.

If you then do a limitedResultTable with 10 as the limit it will show 10 rows and no mention that the test have obfuscated/("limited") the example output.

Hunterness: This is used for `SHOW FUNCTIONS` and `SHOW PROCEDURES` so that we can actually show a table that is not ridiculously huge and unreadable.

We want to be able to show the commands without using `YIELD` and `LIMIT` and due to there being a huge amount of default procedures and functions those lists become very big (and are covered in other parts of the documentation). 

They also have a lot of columns when using `YIELD` so we have one example of each that just lists the columns we talk about in the example and then have a `...` column, https://neo4j.com/docs/cypher-manual/current/clauses/listing-procedures/#_listing_procedures_with_other_filtering.

I hope this explanation helps :)

PR #1235, opened 2021-08-30
Fix wrong link for privileges to show role privileges

Hunterness:
Should link to `SHOW PRIVILEGE` privilege not `ROLE MANAGEMENT` privileges.


martin-neotech (approved):
Build and tested the fix.

PR #1238, opened 2021-08-31
Update constraint syntax

HannesSandberg:
Syntax changed to `FOR ... REQUIRE` instead of `ON ... ASSERT`


HannesSandberg:
Should we add `FOR` & `REQUIRE` to the list of reserved keywords?

Hunterness:
CIP: https://github.com/opencypher/openCypher/pull/166 (and just the updates we did https://docs.google.com/document/d/1LemTgk1hcWQhSlFl9jEsiKciYyBV2yf_XVDgKzt29RI/edit?usp=sharing I believe)

Accompanying manual-modeling PR: https://github.com/neo-technology/neo4j-manual-modeling/pull/2332

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -50,6 +50,31 @@ Replaced by:
...
+
+a|
+label:syntax[]
+label:added[]
+[source, cypher, role="noheader"]
+----
+CREATE CONSTRAINT FOR ... REQUIRE ...
+----
+a|
+New syntax for creating constraints, applicable to all constraint types.
```
Lojjs: Not really sure about this one, but I wonder if we should change the 2 constraint entries just above as well:

1. CREATE CONSTRAINT [constraint_name]... 
Should this have the new FOR .. REQUIRE syntax as well? From the users points of view we will be introducing both changes simultaneously so perhaps it is more helpful with the new syntax there? Or what do you think? 

2. DROP CONSTRAINT...
This one look weird/wrong to me. Wasn't all DROP CONSTRAINT by schema commands deprecated long before 4.4?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
```diff
@@ -184,15 +191,22 @@ class ConstraintsTest extends DocumentingTestBase with SoftReset {
...
-      queryText = "CREATE CONSTRAINT constraint_name ON (book:Book) ASSERT book.isbn IS NOT NULL",
+      queryText = "CREATE CONSTRAINT constraint_name FOR (book:Book) REQUIRE book.isbn IS NOT NULL",
       assertions = _ => assertConstraintWithNameExists("constraint_name", "Book", List("isbn"))
     )
+    testQuery(
+      title = "Create a node property existence constraint using deprecated syntax",
+      text = "The node property existence constraint ensures that all nodes " +
+        "with a certain label have a certain property.",
+      queryText = "CREATE CONSTRAINT ON (book:Book) ASSERT book.title IS NOT NULL",
+      assertions = _ => hasNodeConstraint("Book", "title")
```
Lojjs: For completeness, do you think we should also have examples with the even more deprecated `CREATE CONSTRAINT ON (book:Book) ASSERT exists(book.title)` as well? 

Same for relationship existence constraints.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -50,6 +50,31 @@ Replaced by:
...
+[source, cypher, role="noheader"]
+----
+CREATE CONSTRAINT ON ... ASSERT ...
+----
+a|
+Replaced by:
+[source, cypher, role="noheader"]
+----
+CREATE CONSTRAINT FOR ... REQUIRE ...
+----
```
Lojjs: We should also update the replacements of `CREATE CONSTRAINT ... ASSERT exists()` below to this, so users don't have to migrate in two steps

Lojjs (changes requested):
Overall it looks good, only file I found which you should also consider updating is PrettifierTest.scala. And then I have some smaller thoughts I put in the comments. 

I have built the Cypher manual and confirmed everything looks ok, not sure how to build the refcard these days though.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
HannesSandberg: 1. makes sense.

2. I don't know. Is there a easier way to find out when than using git blame? Trello?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
HannesSandberg: Sure, might be useful if someone want to figure out the difference between new and old syntax while updating.

Hunterness:
> Overall it looks good, only file I found which you should also consider updating is PrettifierTest.scala.

There are three such files, but I think I remember something about those being unused so that's why we skipped them (`PrettifierParser` and `PrettifierParserTest` or something like that) (looks like maybe this was what stopped using it: https://github.com/neo4j/neo4j-documentation/pull/889/)

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
Hunterness: 2. The drop constraint was deprecated but when we added multi-property uniqueness constraints it got enabled for both create and drop, which is why drop is here with both an added and deprecated label

1. If we update the syntax in the create statement I would prefer to have the syntax update before the uniqueness constraint update in the table

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
```diff
@@ -201,6 +201,13 @@ class ConstraintsTest extends DocumentingTestBase with SoftReset {
       queryText = "CREATE CONSTRAINT ON (book:Book) ASSERT book.title IS NOT NULL",
       assertions = _ => hasNodeConstraint("Book", "title")
     )
+    testQuery(
+      title = "Create a node property existence constraint using deprecated syntax",
```
Hunterness: I don't think these can have the same title/name as the other one if we want to show both...

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
HannesSandberg: 3. sounds good

Lojjs (approved):
I'm happy with this now

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
```diff
@@ -59,15 +59,22 @@ class ConstraintsTest extends DocumentingTestBase with SoftReset {
...
       text = "When creating a unique constraint, a name can be provided. The constraint ensures that your database " +
         "will never contain more than one node with a specific label and one property value.",
-      queryText = "CREATE CONSTRAINT constraint_name ON (book:Book) ASSERT book.isbn IS UNIQUE",
+      queryText = "CREATE CONSTRAINT constraint_name FOR (book:Book) REQUIRE book.isbn IS UNIQUE",
       assertions = _ => assertConstraintWithNameExists("constraint_name", "Book", List("isbn"))
     )
+    testQuery(
+      title = "Create a unique constraint using deprecated syntax",
+      text = "The unique constraint ensures that your database " +
+        "will never contain more than one node with a specific label and one property value.",
```
AlexicaWright, line 68:
```suggestion
        "never contains more than one node with a specific label and one property value.",
```
We like to use present tense and active voice, but I am aware that this has not been consistently implemented, and not in this section either. Perhaps it is better to leave it as is, for consistency's sake.

AlexicaWright (approved):
Apart from the comment I left, I think this looks good. It's a fairly straightforward update. 👍

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ConstraintsTest.scala
Lojjs: Yes, I think this is just copy paste from another test in the same file, so perhaps it is better to make a cleanup card on either your Trello board our ours about updating it across the board.

AlexicaWright: I agree, it's best dealt with separately. And with that, I have no more objections. Should I merge?

Hunterness: please do :)

PR #1248, opened 2021-09-06
remove dead code

martin-neotech:
remove the graphgist logic because it is not used.


martin-neotech:
The content in cypher/graphgist/src/test/resources/patterns-in-practice.adoc was used in the Getting Started Manual. 

The content now lives in https://github.com/neo4j/docs-getting-started/blob/dev/modules/ROOT/pages/cypher-intro/patterns-in-practice.adoc

recrwplay (approved):
Looks good - thanks!

PR #1249, opened 2021-09-07
Updated the id function description

renetapopova:
Updates the scalar functions to indicate that id() is unique by object type and database. i.e. it is perfectly allowable for
id() to return the same value for both nodes and relationships in the same database.


martin-neotech (approved):
Thanks

PR #1258, opened 2021-09-15
Escape all default values with `+++` to ensure no formatting is done

Roxling:
This should fix issues with bold/italic text in default values of the setting appendix, e.g `metrics.filter`

This should be cherry-picked forward (4.3 & 4.4)
### NOTE
This affects **all** settings in the appendix. I've tried to verify that it does not add any undesired formatting, but please double check all values before merging!


MarkWoulfeNeo4j (approved):
All present default values appear to be free from bold and italics from what I can see.

There are some default values that are empty though, might want to address them in a separate PR.

PR #1260, opened 2021-09-17
Add examples for failing to create already existing indexes and constraints

Hunterness:
Do we want to add these tests in older versions as well? And if so, how far back?

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-search-performance/index.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-search-performance/index.asciidoc
@@ -252,6 +252,14 @@ include::../indexes/create-a-relationship-type-lookup-index.asciidoc[leveloffset
 
 include::../indexes/create-a-token-lookup-index-specifying-the-index-provider.asciidoc[leveloffset=+1]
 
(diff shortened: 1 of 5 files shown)


Hunterness:
[Can we add the type of exception that will be thrown to 5.4 Constraints?](https://trello.com/c/PywLaToe/2600-can-we-add-the-type-of-exception-that-will-be-thrown-to-54-constraints)

davidoliverSP2 (approved):
This looks good, thanks @Hunterness 

As already discussed, the generated code examples is including a `copy to clipboard` option on the error message, which I don't think is useful (see screenshot example). 

If it's possible to add `role=nocopy` to the codeblock attributes, then that should resolve it, otherwise we can raise another card to investigate.

Screenshot 2021-09-20 at 12 49 40

davidoliverSP2:
Thanks @Hunterness - the last commit was successful in removing the `copy to clipboard` option from the error messages!
Screenshot 2021-09-20 at 16 07 47

nadja-muller (approved):

Hunterness:
@davidoliverSP2 (and @nadja-muller), do we want to backport this to earlier versions as well? If so, from when?

(some syntax needs to be updated to previous versions but should not be hard to backport)

(DECISION: no backporting at this point in time)

PR #1269, opened 2021-10-13
Create instance of OIDCSetting for OpenID Connect docs

mnd999:
GroupSettings are not instantiated when building the configuration
object, thus they are not included in the generated docs. Create an
instance of OIDCSetting so at least the OIDC settings are documented.

In future we should consider doing this for the other groupsettings instances, but they don't all have the required annotations to produce docs, so this is more work.


Lojjs (approved):

davidoliverSP2 (approved):
thanks @mnd999

PR #1272, opened 2021-10-15
Fix tests asserting on plan description

hedengran:
Update tests expected to fail after merging https://github.com/neo-technology/neo4j/pull/12503


tobias-johansson (commented):
👍

PR #1274, opened 2021-10-20
Speeds up the build on macOS

ncordon:
## What
Speeds up the build on OSX systems, by using a file system in memory rather than in disk. The reason for that is that `sun.nio.ch.FileChannelImpl.force` is [10x times](https://bugs.openjdk.java.net/browse/JDK-8205162) slower in an OSX than in Linux. Hat tip to @zimmre.

Writing the database files to memory speeds up that.

## Why
The build was taking about 2 hours on a macOS compared to 20 minutes in Linux. After adding this changes, the build times have lowered to ~20 min in macOS:

Screenshot 2021-10-19 at 19 39 54

### DIFF
--- a/cypher/cypher-docs/src/test/java/org/neo4j/cypher/TestEnterpriseDatabaseManagementServiceBuilder.java
+++ b/cypher/cypher-docs/src/test/java/org/neo4j/cypher/TestEnterpriseDatabaseManagementServiceBuilder.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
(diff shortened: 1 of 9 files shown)


MarkWoulfeNeo4j:
I have tested this and build times are greatly improved:

`mvn clean install` -> 20 mins
`mvn -T1C clean install` -> 10 mins 

🥳

davidoliverSP2:
> I have tested this and build times are greatly improved:
> 
> `mvn clean install` -> 20 mins `mvn -T1C clean install` -> 10 mins
> 
> 🥳

Yep, this works for me too.

@ncordon Thank you very much! Is this PR viable for cherry picking back to earlier versions?

ncordon:
> Is this PR viable for cherry picking back to earlier versions?

Probably, I'll try to give you a hand with that once the new database release is done 💃

martin-neotech (approved):
Thanks

PR #1271, opened 2021-10-14
Document `CALL { ... } IN TRANSACTIONS`

tobias-johansson:
I decided to put this as a subsection of the `CALL {} (subquery)` clause section, rather than as a top level clause section.

PR #1294, opened 2021-10-26
Described `outerTransactionId` of `SHOW TRANSACTIONS` more precisely

arnefischereit:
Depends on #1271

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TransactionsCommandTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TransactionsCommandTest.scala
@@ -92,7 +92,7 @@ class TransactionsCommandTest extends DocumentingTest {
 |m|LONG
 
 |m|outerTransactionId
-|a|The ID of this transaction's outer transaction if such exists. For details, see <<subquery-call-in-transactions, `CALL { ... } IN TRANSACTIONS`>>.
+|a|The ID of this transaction's outer transaction, if such exists. For details, see <<subquery-call-in-transactions, `CALL { ... } IN TRANSACTIONS`>>.
 |m|STRING
 
 |m|metaData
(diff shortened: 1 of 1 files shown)


hedengran (approved):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TransactionsCommandTest.scala
```diff
@@ -92,7 +92,7 @@ class TransactionsCommandTest extends DocumentingTest {
 |m|LONG
 
 |m|outerTransactionId
-|a|The ID of an outer transaction.
+|a|The ID of this transaction's outer transaction if such exists. For details, see <<subquery-call-in-transactions, `CALL { ... } IN TRANSACTIONS`>>.
```
AlexicaWright:
```suggestion
|a|The ID of this transaction's outer transaction, if such exists. For details, see <<subquery-call-in-transactions, `CALL { ... } IN TRANSACTIONS`>>.
```

AlexicaWright (commented):
Just a small punctuation comment.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/TransactionsCommandTest.scala
arnefischereit: Interesting. I thought that one is supposed to leave that comma out[,] if the conditional clause is the second clause, because the `if` already signifies the clause boundary.
See https://english.stackexchange.com/a/28625

AlexicaWright: Bear in mind that this is not an ordinary sentence, it is part of a table and as such does not adhere to the same rules. But if you feel strongly about it, feel free to leave it out. It is not incorrect.

PR #1303, opened 2021-11-09
Clarify that the password is a string and not a plain token as the username

Hunterness:
Also correct wrong index syntax...

PR #1305, opened 2021-11-12
Clarify that the password is a string (#1303)

martin-neotech:
Clarify that the password is a string and not a plain token as the username.

Also correct wrong index syntax.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
@@ -3,40 +3,55 @@
 |===
 | Command | Description | Type of user | Available in Community Edition
 
-| [source, cypher, role=noplay]
+|
+[source, cypher, role=noplay]
+----
 SHOW USERS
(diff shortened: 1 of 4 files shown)


cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
```diff
@@ -3,40 +3,55 @@
 |===
 | Command | Description | Type of user | Available in Community Edition
 
-| [source, cypher, role=noplay]
+|
+[source, cypher, role=noplay]
+----
```
Hunterness, line 8: Not sure about what these changes are but I guess those are updates from you?

Hunterness (approved):

cypher/cypher-docs/src/docs/dev/ql/administration/security/user-management-syntax.asciidoc
martin-neotech: Yes, it is just minor fixes I did, so it is correct asciidoc :)

PR #1311, opened 2021-11-22
Update CREATE FULLTEXT syntax summary to follow the rules we set up for such summaries

Hunterness:


Hunterness:
[CREATE FULLTEXT INDEX includes | which is also used in syntax summaries for OR, should be escaped](https://trello.com/c/I8cYmdax/2699-create-fulltext-index-includes-which-is-also-used-in-syntax-summaries-for-or-should-be-escaped)

HannesSandberg (approved):
lgtm 👍

cypher/cypher-docs/src/docs/dev/ql/administration/indexes-for-full-text-search/create-fulltext-syntax.asciidoc
```diff
@@ -6,7 +6,7 @@
 | [source, cypher, role=noplay]
 ----
 CREATE FULLTEXT INDEX [index_name] [IF NOT EXISTS]
-FOR (n:LabelName[\|...])
+FOR (n:LabelName["\|" ...])
```
AlexicaWright, line 9: Why is it `index_name` here but just `name` above? Also, why only one pipe here when two pipes in the example above?

AlexicaWright (commented):
A few questions.

AlexicaWright (approved):

PR #1313, opened 2021-11-24
Pagecache mem setting is a long

tinwelint:
Follows a setting type change in neo4j dev branch


AlexicaWright:
Builds fine locally.

PR #1321, opened 2021-11-25
coverted long value to needed string value

glindroth:


tinwelint (approved):

davidoliverSP2:
Thanks @glindroth - do you know where does this change render in the docs (so I can test that it's updating correctly)

Also, could you confirm this is only required for version 5.0?

tinwelint (changes requested):
The changes in Neo4jBasicDocTest and StartWithConfigurationDocTest are not correct. They should stay Long. It happens to be correct locally perhaps since you don't compile against latest neo4j HEAD

glindroth:
> The changes in Neo4jBasicDocTest and StartWithConfigurationDocTest are not correct. They should stay Long. It happens to be correct locally perhaps since you don't compile against latest neo4j HEAD

I have reverted those changes. With that revert this builds locally at least.

tinwelint (approved):

davidoliverSP2 (approved):
👍 

thanks @glindroth & @tinwelint

PR #1325, opened 2021-11-26
Fix broken refcard test by loosing up the assertion

Hunterness:
Should fix: https://trello.com/c/049nLin1/2740-test-failure-showprocfunctestproducedocumentation-is-failing-consistently-on-dev-docstest-build


mnd999 (approved):
LGTM, have built it as well and it seems fine.

Weirdly, the .pom files in docs on dev are 4.4.0-SNAPSHOT still, but I don't think that's a problem for this PR.

davidoliverSP2 (approved):
Thanks @Hunterness

PR #1327, opened 2021-12-02
Remove use of DROP INDEX/CONSTRAINT by schema

Hunterness:
The command is being removed in https://github.com/neo-technology/neo4j/pull/13438 (but this pr is not dependent on that going in first)


ncordon (approved):

davidoliverSP2 (approved):
Thanks @Hunterness - I can confirm this removes the sections highlighted. I'll have to assume that the technical review covered whether these are the _right_ sections 😉

PR #1366, opened 2022-01-14
The file name for load csv needs to be url encoded if it contains spaces

renetapopova:


cypher/cypher-docs/src/docs/dev/ql/load-csv/index.asciidoc
```diff
@@ -51,6 +51,7 @@ This is definitely not recommended.
 File URLs will be resolved relative to the `dbms.directories.import` directory.
 For example, a file URL will typically look like `+file:///myfile.csv+` or `+file:///myproject/myfile.csv+`.
 
+* When using `+file:///+` URLs, if the file name has spaces or other than alpha numerics, then it needs to be URL encoded ( see https://www.w3schools.com/tags/ref_urlencode.ASP).
```
arnefischereit: Should we possibly rather link to something more institutional like Wikipedia instead of some arbitrary online school?
```suggestion
* When using `+file:///+` URLs, if the file name has spaces or other than alpha numerics, then it needs to be URL encoded (see https://en.wikipedia.org/wiki/Percent-encoding).
```

Also:

The way the sentence is phrased, I would expect that one possibly has to rewrite the whole URL once some non-alpha-numerical character appears. Instead one would "only" need to encode the special characters in percent encoding.

```suggestion
* When using `+file:///+` URLs, spaces and other non-alphanumeric characters need to be URL encoded (see https://en.wikipedia.org/wiki/Percent-encoding).
```

arnefischereit (approved):
Change is correct but I have raised two improvement suggestions.

cypher/cypher-docs/src/docs/dev/ql/load-csv/index.asciidoc
renetapopova: Thanks for the suggestions, @arnefischereit! I'll fix the wording. Regarding Wikipedia, I might disagree with you. It's a public space, where everybody can write whatever they want or delete whenever they like. I tried to find a better source than w3c school, so so far this is the one: https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding.

arnefischereit: @renetapopova: I am totally fine with the Mozilla knowledge base, as I would deem them an authority in that space as well.

renetapopova: Excellent! Then, I'll change it to the Mozilla link. Thanks, @arnefischereit!

PR #1367, opened 2022-01-14
Refers operators from Indexes for search performance

renetapopova:


Hunterness (approved):
I haven't built and checked that the link works but otherwise it looks fine :)

PR #1419, opened 2022-02-03
Adds an example of how to use CASE WHEN  as part of WITH

renetapopova:


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
```diff
@@ -39,8 +39,11 @@ class CaseExpressionsTest extends DocumentingTest {
                   #  (bob)-[:KNOWS]->(daniel),
                   #  (charlie)-[:KNOWS]->(daniel),
                   #  (bob)-[:MARRIED]->(eskil)""".stripMargin('#'))
-    p("""Generic conditional expressions may be expressed using the well-known `CASE` construct.
+   p("""Generic conditional expressions may be expressed using the `CASE` construct.
```
arnefischereit: I think the previous indentation worked better.
```suggestion
    p("""Generic conditional expressions may be expressed using the `CASE` construct.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
```diff
@@ -128,5 +131,23 @@ class CaseExpressionsTest extends DocumentingTest {
         resultTable()
       }
     }
+    section("Using the result of `CASE` in the succeeding clause or statement", "syntax-use-case-result") {
+      p("""You can use the result of `CASE` to set a property(s) or a label(s) on a node or relationship.
+       #For example, instead of specifying the node directly, you can set a property for a node selected by an expression:""".stripMargin('#'))
```
arnefischereit: Indentation:
```suggestion
          #For example, instead of specifying the node directly, you can set a property for a node selected by an expression:""".stripMargin('#'))
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
```diff
@@ -128,5 +131,23 @@ class CaseExpressionsTest extends DocumentingTest {
         resultTable()
       }
     }
+    section("Using the result of `CASE` in the succeeding clause or statement", "syntax-use-case-result") {
+      p("""You can use the result of `CASE` to set a property(s) or a label(s) on a node or relationship.
```
arnefischereit: I would rather use the plural form all the way:
```suggestion
      p("""You can use the result of `CASE` to set properties on a node or relationship.
```
and I find it particular that you pick the labels: We cannot bind a label to a variable as far as I know and if we bind the node/relationship to a variable, then we could do practically anything with that and not just assign a label.

Do you have a good example for that, that you could add?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
```diff
@@ -128,5 +131,23 @@ class CaseExpressionsTest extends DocumentingTest {
...
+              #  WHEN 'blue'  THEN 1
+              #  WHEN 'brown' THEN 2
+              #  ELSE 3
+              #END AS colourCode
+              #SET n.colourCode = colourCode""".stripMargin('#'),
+      ResultAssertions((r) => {
+            assertStats(r, propertiesWritten = 5)
+          })) {
+          p("For more information about using the `SET` clause, see <<query-set>>.")
+          resultTable()
```
arnefischereit: Indentation
```suggestion
        ResultAssertions((r) => {
          assertStats(r, propertiesWritten = 5)
        })) {
        p("For more information about using the `SET` clause, see <<query-set>>.")
        resultTable()
```

arnefischereit (changes requested):
I think we should indent differently and maybe you could remove the reference to the labels or provide an example what you mean there.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
renetapopova: I don't have. I'll remove the label's part.

arnefischereit (approved):
LGTM

PR #1436, opened 2022-02-14
Use the Neo4j Prettifier instead of a custom one.

Lojjs:
As the parboiled parser is being removed, I could see three options:

1. Change the docs PrettifierParser to be implemented on top of som Scala library like RegexParser instead. I played around with this for like a day but backed out as it was super complex.

2. Use the Neo4j prettifier instead of a custom one (that is the approach taken in this PR).

3. Not have any automatic prettifying in docs, but just prettify queries manually at need.

As well as review the code changes, also think about if option 2 was the right pick.

Needs to be merged together with/before https://github.com/neo-technology/neo4j/pull/14320


Lojjs:
[Remove parboiled parser](https://trello.com/c/qi8KrxPK/2793-remove-parboiled-parser)

ncordon (approved):
Looks good to me. Also I think it's the right choice from the 3 you listed!

martin-neotech (approved):

martin-neotech:
Lets wait until https://github.com/neo4j/neo4j-documentation/pull/1437 has been merged, this one fixes the broken build.

PR #1438, opened 2022-02-21
fix build error

martin-neotech:


Lojjs (approved):
Seems to match Neo4j version, not sure why it didn't break until now though

PR #1447, opened 2022-03-03
Await databases to start when starting dbms (flaky test fix).

tselmegbaasan:
Recently dbms start was changed. It is now asynchronous, which means that non-system databases are not guaranteed to have started when dbms is done starting, but they will eventually start. This await mechanism ensures that a specified database is started.


martin-neotech (approved):
Thanks

PR #1451, opened 2022-03-15
Remove support for exists(property)

Lojjs:
This has been replaced by `property IS NOT NULL`, see https://github.com/neo-technology/neo4j/pull/14699

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
@@ -119,6 +119,11 @@ class PredicateFunctionsTest extends DocumentingTest {
         p("The names of all nodes with the `name` property are returned, along with a boolean (`true` or `false`) indicating if they are married.")
         resultTable()
       }
+      note {
+        p(
+          """Note that the **function** `exists()` looks very similar to the **clause** `EXISTS {...}`, but they are not the same and should not be confused.
+            #See <<existential-subqueries, Using existential subqueries in `WHERE`>> for more information.""".stripMargin('#'))
+      }
     }
     section("isEmpty()", "functions-isempty") {
       p("The function `isEmpty()` returns `true` if the given list or map contains no elements or if the given string contains no characters.")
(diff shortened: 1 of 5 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -98,13 +98,13 @@ class PredicateFunctionsTest extends DocumentingTest {
       }
     }
     section("exists()", "functions-exists") {
-      p("""The function `exists()` returns `true` if a match for the given pattern exists in the graph, or if the specified property exists in the node, relationship or map.
+      p("""The function `exists()` returns `true` if a match for the given pattern exists in the graph.
           #`null` is returned if the input argument is `null`.""".stripMargin('#'))
```
AzuObs, line 102: I think that perhaps exists should now really be either an expression or a clause. In the code I've seen it called ExistsSubqueryClause although it extends Expression, which is ambiguous. I've not seen it called a Function though. In the docs we are treating the existential subquery and the function the same way but that is probably a mistake.

The other point I would like to raise is that exists() is not valid syntax for existential subqueries. It needs curly braces instead, eg exists { } and will produce strange parsing exceptions if it doesn't use those.

Lojjs: Existential subqueries is documented in another place. This is now only for `exists(pattern)`, which I agree is not really a function even if it happens to be implemented as one. I will ask someone from docs team to review this and then they can maybe come with a concrete suggestion on how to rephrase/move this part of documentation to make it clearer

AzuObs (approved):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
AlexicaWright: I agree that it may be confusing with both the function `exists()` and the existential subquery `EXISTS {}`, but whether this should or shouldn't be is beyond me. If it is implemented as a function, it belongs in this section. 
If you feel the risk of confusion between the two is too great, you could put a note about it in the Predicate functions section, not in the glossary.

AlexicaWright (approved):
👍  I think this clarifies the possible confusion. Thank you @Lojjs !

PR #1477, opened 2022-04-20
Introduce isNaN()

AzuObs:
Adding the isNaN function which returns true if the given number is `NaN`. Co-authored with @nadja-muller

### DIFF
--- a/cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/MathematicalFunctionsTest.scala
+++ b/cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/MathematicalFunctionsTest.scala
@@ -61,6 +61,14 @@ The absolute value.
 ###assertion=returns-one parameters=expression
 RETURN
 
+isNaN($expr)
+###
+
+Returns whether a number is `NaN`.
+
+###assertion=returns-one parameters=expression
+RETURN
+
 rand()
 ###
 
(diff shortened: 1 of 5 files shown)


cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/MathematicalFunctionsTest.scala
```diff
@@ -61,6 +61,14 @@ The absolute value.
...
 RETURN
 
+isNaN($expr)
+###
+
+Returns whether a number is `NaN`.
+
+###assertion=returns-one parameters=expression
+RETURN
+
```
Lojjs, line 71: I know @martin-neotech is working on a replacement for the refcard, so leaving this comment here for visibility, this should be added to the 5.0 version of the cheat sheet

Lojjs (approved):
Great work! I built it locally together with your Neo4j PR and verified that all links and rendering works as expected.

AlexicaWright (approved):
Looks good to me, content-wise, but as the tests are failing I haven't been able to build and test it locally.

AzuObs:
> Looks good to me, content-wise, but as the tests are failing I haven't been able to build and test it locally.

Thank you for the approve 😸 I will fix the tests before merging this PR. I should have told you but the reason they are failing is because the code doesn't exist in the main database yet.

PR #1490, opened 2022-05-17
Adds docs for COUNT expression

ncordon:
Adds documentation for `COUNT` expression, which substitutes `size` for certain uses: `size((a)-[r]->(b))`

## TODO
- [ ] Differences between `COUNT` and `EXISTS`
- [ ] Optimizations made for `COUNT` internally?
- [ ] `COUNT` as groupping key
- [ ] Double check `size((a)-[r]->(b))` has completely been removed from the docs


cypher/cypher-docs/src/docs/dev/clause.asciidoc
```diff
@@ -313,6 +313,10 @@ include::ql/query-unwind.adoc[leveloffset=+1]
 //Reading sub-clauses
 include::ql/query-where.adoc[leveloffset=+1]
 
+include::ql/query-exists.adoc[leveloffset=+1]
+
+include::ql/query-count.adoc[leveloffset=+1]
```
Lojjs, line 322: The new sections does not show up when I build, I think there is something in manual-modelling or antora repo that is needed as well. The docs team should know for sure

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    p("""#It can be used in the same way as a path pattern but it allows you to use `WHERE` clauses internally.
+        #A subquery has a scope, as indicated by the opening and closing braces, `{` and `}`.
+        #Any variable that is defined in the outside scope can be referenced inside the subquery's own scope.
+        #Variables introduced inside the subquery are not part of the outside scope and therefore can't be accessed on the outside.""".stripMargin('#'))
+    functionWithCypherStyleFormatting("""COUNT {
+                                        #  [Pattern]
+                                        #  WHERE [Expression]
+                                        #}""".stripMargin('#'))
+    p("The `WHERE` clause is optional.")
+    p("It is worth noting that the `MATCH` keyword can be omitted in subqueries and that the `WHERE` clause is optional.")
```
Lojjs: We should remove this line right? As the WHERE part is covered in the previous sentence and the MATCH part isn't true for COUNT, only for EXISTS

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
```diff
@@ -0,0 +1,163 @@
...
+    section("Using count subqueries in `WHERE`", "count-subqueries") {
+      section("Simple count subquery", "count-subquery-simple-case") {
+        p("""Variables introduced by the outside scope can be used in the inner `MATCH` clause. The following query exemplifies this and would output the owners of more tahan one dog:""")
```
Lojjs:
```suggestion
        p("""Variables introduced by the outside scope can be used in the inner `MATCH` clause. The following query exemplifies this and would output the owners of more than one dog:""")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
```diff
@@ -0,0 +1,163 @@
...
+                #  WHERE person.name = dog.name
+                #} > 0
+                #RETURN person.name AS name""".stripMargin('#'),
+          ResultAssertions(r => {
+            r.toList should equal(List(Map("name" -> "Andy")))
+          })) {
+          resultTable()
+        }
+      }
+      section("Nesting count subqueries", "count-subquery-nesting") {
```
Lojjs, line 88: We don't have any test for nesting in the Neo4j PR, right?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
```diff
@@ -0,0 +1,163 @@
...
+        }
+      }
+      section("Count subquery with `WHERE` clause", "count-subquery-with-where") {
+        p("""A `WHERE` clause can be used inside the `COUNT` pattern.
+            #Variables introduced by the `MATCH` clause and the outside scope can be used in this scope.""".stripMargin('#'))
+        query("""MATCH (person:Person)
+                #WHERE COUNT {
+                #  (person)-[:HAS_DOG]->(dog:Dog)
+                #  WHERE person.name = dog.name
+                #} > 0
```
Lojjs, line 80: I was considering if it would be interesting to mention that `COUNT > 0`is semantically equivalent to  `EXISTS`, but to not confuse people we should maybe wait with that until we have improved both so they work in the same places.

Lojjs (changes requested):
The text and examples are great, just some minor comments that I posted below. 

I think there are still a few more things we would like to cover as well, but not sure whether we want to wait for the other COUNT/EXISTS work first or not:

1. ~Add COUNT to the [clause glossary](https://github.com/neo4j/neo4j-documentation/blob/dev/cypher/cypher-docs/src/docs/dev/glossary.asciidoc)~
2. ~Add COUNT under 5.0 and New features [here](https://github.com/neo4j/neo4j-documentation/blob/dev/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc)~
3. Add an example of COUNT on a node pattern and write a note in it about that it does not support a label expression, properties or a WHERE clause
4. Possibly mention that you can use this instead of the more cumbersome `size([x=(a)-[r]->(b) | x])` (not sure if we want to included this in docs or not)
5. Possibly mention the limitations around named patterns and multiple pattern parts (not sure if we want to included this in docs or not)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
ncordon: Yes, although I was wondering whether we should bring back this minor adjustment to the COUNT. The reason I was thinking of is because one should expect to be able to substitute a `COUNT > 0` by an `EXISTS` and viceversa, and they should have same syntax. So either we remove it from the `EXISTS` or we add it to the `COUNT` to align both.

cypher/cypher-docs/src/docs/dev/clause.asciidoc
ncordon: Good catch. I've fixed that in here: https://github.com/neo-technology/neo4j-manual-modeling/pull/2859

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CountTest.scala
Lojjs: Good point, we should probably align them and removing it from `EXISTS` would require a deprecation. As it is just syntactic sugar, I think it only require works on the surface side, but it might still be worth asking Satia if he has any input on it

cypher/cypher-docs/src/docs/dev/clause.asciidoc
Lojjs: Nice, I approved that one, although we should wait to merge it ofc

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -521,6 +521,16 @@ isNaN(n.prop)
...
 
+a|
+label:functionality[]
+label:added[]
+[source, cypher, role="noheader"]
+----
+COUNT { (n) WHERE n.foo = "bar" }
+----
+a|
+New expression which returns whether the number of elements of a pattern.
```
Lojjs, line 532:
```suggestion
New expression which returns the number of occurrences of a pattern.
```

That might be better?

cypher/cypher-docs/src/docs/dev/glossary.asciidoc
```diff
@@ -73,6 +73,8 @@ This section comprises a glossary of all the keywords -- grouped by category and
 |<<query-with, WITH ... [AS]>>                        | Projecting   |  Allows query parts to be chained together, piping the results from one to be used as starting points or criteria in the next.
 |<<query-where, WHERE>>                          | Reading sub-clause | A sub-clause used to add constraints to the patterns in a `MATCH` or `OPTIONAL MATCH` clause, or to filter the results of a `WITH` clause.
 |<<existential-subqueries, WHERE EXISTS {...}>>  | Reading sub-clause | An existential sub-query used to filter the results of a `MATCH`, `OPTIONAL MATCH` or `WITH` clause.
+|<<count-subqueries, COUNT {...}>>  | Reading sub-clause | An expression used to compute the number of results of a subquery inside another clause, e.g. `MATCH`, `WITH`, `SET`, etc.
```
Lojjs, line 76: This link goes to the  `Using count subqueries in WHERE` , perhaps we want to send users to the top of the page instead?

davidoliverSP2:
[Document COUNT {} expression](https://trello.com/c/fcTIa9ue/4236-document-count-expression)

sherfert:
> 3. Add an example of COUNT on a node pattern and write a note in it about that it does not support a label expression, properties or a WHERE clause
> 4. Possibly mention that you can use this instead of the more cumbersome `size([x=(a)-[r]->(b) | x])` (not sure if we want to included this in docs or not)
> 5. Possibly mention the limitations around named patterns and multiple pattern parts (not sure if we want to included this in docs or not)

These are not relevant any longer.

sherfert:
Replaced by #1590

PR #1504, opened 2022-06-07
Adapted URL in test to use https to avoid redirect

arnefischereit:
Before the test would fail because we would get a 301 redirect instead
of the desired csv file.


davidoliverSP2 (approved):
thanks @arnefischereit

PR #1505, opened 2022-06-08
Adapted URL in test to use https to avoid redirect

davidoliverSP2:
Before the test would fail because we would get a 301 redirect instead
of the desired csv file.

Cherry picks #1504


davidoliverSP2:
Duplicated in #1511

PR #1524, opened 2022-06-21
Add documentation to explain how NULL works in a CASE expression

gem-neo4j:


gem-neo4j:
[CASE WHEN null doesn't work properly](https://trello.com/c/Svugko7N/102-case-when-null-doesnt-work-properly)

AzuObs (approved):

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/CaseExpressionsTest.scala
```diff
@@ -149,5 +149,22 @@ class CaseExpressionsTest extends DocumentingTest with QueryStatisticsTestSuppor
...
     }
+    section("Using `CASE` with NULL values", "syntax-use-case-with-null") {
+      p("""When using the simple `CASE` form, it is useful to remember that in Cypher `null = null` yields `null`.
+          #For example, you might expect `age_10_years_ago` to be `-1` for the node named `Daniel`:""".stripMargin('#'))
+      query("""MATCH (n)
+              #RETURN n.name,
+              #CASE n.age
+              #  WHEN NULL THEN -1
+              #  ELSE n.age - 10
+              #END AS age_10_years_ago""".stripMargin('#'),
```
AzuObs: Fairly minor, but I would suggest taking a closer look at the query and maybe aligning more with the style guide https://neo4j.com/developer/cypher/style-guide/#cypher-null-boolean. 

It looked a bit unusual when I read it, so I checked. And it seems to me that `WHEN NULL` should be `WHEN null`

lidiazuin (approved):

PR #1525, opened 2022-06-22
MixedEditionModule is renamed to EnterpriseEM

gfx54b:


martin-neotech (approved):
Just wasted some time figuring out why the code didnt compile.

So the name was changed again :) (about 18 hour ago).

PR #1537, opened 2022-06-28
Update the constraints for creating a Duration.

JonasHolgersson:
Numeric values 'n' are parsed as a java Long which is a 64-bit integer and not arbitrarily large.
Make it clear that there is an upper bound to how long Duration values can be created.


lidiazuin (approved):

JonasHolgersson:
[Unit-based duration format behaviour differs from documentation](https://trello.com/c/t2Hq30Pw/3584-unit-based-duration-format-behaviour-differs-from-documentation)

PR #1298, opened 2021-11-01
Cypher: document relationship pattern predicates

blouerat:
Note that this depends on https://github.com/neo-technology/neo4j/pull/12784

PR #1544, opened 2022-07-08
Flakiness in tests for triadic operators

blouerat:
I had a couple failures on [my latest PR](https://github.com/neo4j/neo4j-documentation/pull/1298), due to what looks like flaky tests for triadic operators.
Added a bunch of friendships to make the world a better place and to nudge the planner in the right direction.
I've also fixed the setup query so that it uses `stripMargin` properly.


davidoliverSP2:
Hi @blouerat - did you want anyone in the planner team to review this?

From a docs-team & output point of view, is there anywhere in the docs I can verify this, or is it purely for performance?

blouerat:
Hi @davidoliverSP2, sorry my PR description wasn't clear.
This is changing the fixtures for the `QueryPlanTest` class, which is what generates content for the _Execution plan operators in detail_ page if my understanding is correct.
This doesn't make any change to the content itself, except for some _estimated rows_ in some places I would assume.
The motivation for this PR is that currently the tests for _triadic selection/build/filter_ is flaky, the planner doesn't always choose the plan we want to display, and so the build goes red. I've extracted the test logic, analysed it separately with some help from @inmost-light, and we found that by simply adding some `:FRIENDS_WITH` relationships we can nudge the planner in the right direction.
In summarry: will produce the same content, more reliably. 🤖 

As for the added `|` in the query, that's for [`stripMargin`](https://www.scala-lang.org/api/current/scala/collection/StringOps.html#stripMargin:String) to work correctly. It strips off all the spaces and the pipe character for each line, allowing you to indent your multi-line string without polluting it with extra spaces:

```scala
// Properly removes the indetation, no spaces left before BAR and BAZ:
"""FOO
  |BAR
  |BAZ""".stripMargin

/*
FOO
BAR
BAZ
*/

// This is what's in the code currently, the output is also indented:
"""FOO
   BAR
   BAZ""".stripMargin

/*
FOO
   BAR
   BAZ
*/

// The previous example is equivalent to not using stripMargin at all:
"""FOO
   BAR
   BAZ"""

/*
FOO
   BAR
   BAZ
*/
```

Cypher doesn't care for those spaces, so it will behave exactly the same, but it makes debugging a tad nicer.

Hopefully this is now clear enough for you to review, and I'll also ask someone from the planner team to have a look.

davidoliverSP2:
@blouerat many thanks for the explanation! sounds good

Dunfluff (approved):
LGTM

PR #1553, opened 2022-07-18
Fix broken build -> tests not finding database neo4j

Hunterness:
A setting for the database behaviour was updated in https://github.com/neo4j/neo4j-documentation/pull/1530, but was needed in 2 more places in addition to where it was added.

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/DocumentingTestBase.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/DocumentingTestBase.scala
@@ -25,6 +25,9 @@ import org.junit.After
 import org.junit.Before
 import org.neo4j.configuration.GraphDatabaseSettings.DEFAULT_DATABASE_NAME
 import org.neo4j.configuration.helpers.SocketAddress
+import org.neo4j.cypher.ExecutionEngineHelper
+import org.neo4j.cypher.GraphIcing
+import org.neo4j.cypher.TestEnterpriseDatabaseManagementServiceBuilder
 import org.neo4j.cypher.docgen.tooling.DocsExecutionResult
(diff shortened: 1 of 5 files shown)


martin-neotech:
```
org.neo4j.cypher.docgen.MapsTest.2: MATCH {actor:Person {name: 'Charlie Sheen'}}-[:ACTED_IN]->{movie:Movie} RETURN actor{.name, .realName, movies: collect{movie{.title, .year}}}

org.neo4j.exceptions.SyntaxException: Aggregation column contains implicit grouping expressions. For example, in 'RETURN n.a, n.a + n.b + count(*)' the aggregation expression 'n.a + n.b + count(*)' includes the implicit grouping key 'n.b'. It may be possible to rewrite the query by extracting these grouping/aggregation expressions into a preceding WITH clause. Illegal expression(s): actor, name, realName (line 2, column 8 (offset: 79))
      "RETURN actor{.name, .realName, movies: collect(movie{.title, .year})}"
              ^

org.neo4j.exceptions.SyntaxException: 
      Aggregation column contains implicit grouping expressions. For example, in 'RETURN n.a, n.a + n.b + count(*)' the aggregation expression 'n.a + n.b + count(*)' includes the implicit grouping key 'n.b'. It may be possible to rewrite the query by extracting these grouping/aggregation expressions into a preceding WITH clause. Illegal expression(s): actor, name, realName (line 2, column 8 (offset: 79))
      "RETURN actor{.name, .realName, movies: collect(movie{.title, .year})}"
```

Hunterness:
@martin-neotech Yeah, that is a separate issue that became visible when the test actually got to run. Looks like the runtime team might have removed the support for implicit grouping keys (was deprecated) but that documentation hasn't been updated.

martin-neotech (approved):
Lets merge this fix and then fix the next issue.

PR #1554, opened 2022-07-20
update test to not use ambiguous aggregation expression

LinneaAndersson:


martin-neotech:
```
org.neo4j.cypher.docgen.MapsTest.2: MATCH {actor:Person {name: 'Charlie Sheen'}}-[:ACTED_IN]->{movie:Movie} WITH actor.name, actor.realName, collect{movie{.title, .year}} AS movies RETURN actor{.name, .realName, movies: movies}

org.neo4j.exceptions.SyntaxException: Expression in WITH must be aliased (use AS) (line 2, column 6 (offset: 77))
      "WITH actor.name, actor.realName, collect(movie{.title, .year}) AS movies"
            ^
```

LinneaAndersson:
I noticed, I have now updated the test to the best solution I could come up with. Do we need to update the description as well @martin-neotech ?

martin-neotech (approved):
Thanks

PR #1555, opened 2022-07-21
Deprecate case matching on null in 4.4

gem-neo4j:


gem-neo4j:
[CASE WHEN null DEPRECATION](https://trello.com/c/VLgeS0J6/213-case-when-null-deprecation)

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -183,6 +183,32 @@ Please instead use:
...
+MATCH (n)
+RETURN
+CASE n.prop
+    WHEN null THEN 'one'
+    ELSE 'two'
+END
+----
+Currently, if n.prop is null, 'one' would be returned.
+a|
+The matching of null == null in CASE expressions is deprecated and in future versions will no longer be able to be matched on.
```
nadja-muller: - == actually doesn't exist in Cypher, so it should be one =, I believe
- It might be worth to point out that a Cypher comparision would return false here and we want to be consistent.
```suggestion
Since null = anything returns false in Cypher, a when exression will no longer match in future versions.
```

nadja-muller (changes requested):
The == bothers me a bit. But please double check if I'm right.

nadja-muller (approved):

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -183,6 +183,32 @@ Please instead use:
...
+Please use `IS NULL` instead.
+[source, cypher, role="noheader"]
+----
+MATCH (n)
+RETURN
+CASE
+    WHEN n.prop IS NULL THEN 'one'
+    ELSE 'two'
+END
+----
```
davidoliverSP2:
```suggestion
a|
label:functionality[]
label:deprecated[]
[source, cypher, role="noheader"]
----
MATCH (n)
RETURN
CASE n.prop
    WHEN null THEN 'one'
    ELSE 'two'
END
----
a|
Currently, if `n.prop` is `null`, `'one'` would be returned.
Since `null = null` returns `false` in Cypher, a `WHEN` expression will no longer match in future versions.

Please use `IS NULL` instead:
[source, cypher, role="noheader"]
----
MATCH (n)
RETURN
CASE
    WHEN n.prop IS NULL THEN 'one'
    ELSE 'two'
END
----
```

davidoliverSP2 (approved):
Thanks @gem-neo4j 

I've slightly reordered the sentences. I'll commit and merge shortly.

I assume that, as this is a deprecation, no action is required in the 5.0 branch?

davidoliverSP2:
> Thanks @gem-neo4j
> 
> I've slightly reordered the sentences. I'll commit and merge shortly.
> 
> I assume that, as this is a deprecation, no action is required in the 5.0 branch?

Ah I see #1556 covers this in the 5.0 branch 😄

PR #1563, opened 2022-08-01
Update LOAD CSV remote url to https

recrwplay:
This is the same fix that was applied to 4.4 in #1504


davidoliverSP2 (approved):

martin-neotech (approved):
Verified locally.

PR #1571, opened 2022-08-08
Cypher manual - add section about aggregation grouping keys

LinneaAndersson:
Screenshot 2022-09-14 at 09 31 34
Screenshot 2022-09-14 at 09 31 47
Screenshot 2022-09-14 at 09 31 52
Screenshot 2022-09-14 at 09 32 04

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
@@ -69,6 +69,14 @@ class AggregatingFunctionsTest extends DocumentingTest {
         #The latter, `+count(*)+` is an aggregate expression.
         #The matching paths will be divided into different buckets, depending on the grouping key.
         #The aggregate function will then be run on these buckets, calculating an aggregate value per bucket.""".stripMargin('#'))
+    p("""An aggregation functions input expression can contain non-grouping variables, for example `+sum(1 + n.x)+`.
+        #However, it is not possible to mix non-grouping variables and aggregation functions.
+        #The example below will throw an error since we mix `n.x`, which is not a grouping key, with the aggregation expression `+count(*)+`.
+        #For more information see <<grouping-keys,Grouping keys>>.""".stripMargin('#'))
+    p("""[source, cypher]
+        #----
+        #RETURN n.x + count(*)
+        #----""".stripMargin('#'))
     p("To use aggregations to sort the result set, the aggregation must be included in the `RETURN` to be used in the `ORDER BY`.")
     p("""The `DISTINCT` operator works in conjunction with aggregation.
         #It is used to make all values unique before running them through an aggregate function.
@@ -384,6 +392,67 @@ class AggregatingFunctionsTest extends DocumentingTest {
       })) {
         p("The sum of the two supplied Durations is returned.")
         resultTable()
+      }
+    }
+    section("Grouping keys", "grouping-keys") {
+      p("""The expressions which does not contain any aggregation function are the grouping keys of the aggregation.
+          #If there are no grouping keys, the aggregation will be over all rows.
+          #For expressions that contain an aggregation function, the leaves of the expressions must be one of:
+          #
+          #* An aggregation
+          #* A literal
+          #* A Parameter
+          #* A variable - ONLY IF that variable is also one of:
+          #** A projection expression on its own (e.g. the `n` in `RETURN n AS myNode, n.value + count(*)`)
+          #** A local variable in the expression (e.g the `x` in `RETURN n, n.prop + size([ x IN range(1, 10) | x ])`)
+          #* Property access - ONLY IF that property access is also a projection expression on its own (e.g. the `n.prop` in `RETURN n.prop, n.prop + count(*)`)
+          #* Map access - ONLY IF that map access is also a projection expression on its own (e.g. the `map.prop` in `WITH {prop: 2} AS map RETURN map.prop, map.prop + count(*)`)""".stripMargin(
+          '#'))
+      note {
+        p("The same rules apply to expressions in ORDER BY if there is an aggregation in preceding WITH or RETURN clause.")
+      }
+      section("Examples of expressions which mix aggregation and non-aggregation expressions.", "grouping-key-examples") {
+        p("Using non-grouping variables inside an aggregation.")
+        query("""MATCH (p: Person) RETURN max(p.age)""",
+          ResultAssertions((r) => {
+            r.toList.head("max(p.age)") should equal(44L)
+          })) {
+          resultTable()
+        }
+        p("Mix of the aggregation function `max` and the literal `1`.")
+        query("""MATCH (p: Person) RETURN 1 + max(p.age)""",
+          ResultAssertions((r) => {
+            r.toList.head("1 + max(p.age)") should equal(45L)
+          })) {
+          resultTable()
+        }
+        p("Using a variable `n`, which is also a grouping key.")
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n, n.age - max(f.age)""",
+          ResultAssertions((r) => {
+            r.toList.size shouldBe 1
+            r.head.get("n.age - max(f.age)") should equal(Some(-31L))
+          })) {
+          resultTable()
+        }
+        p("Using the property `n.age`, which is also a grouping key.")
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age, n.age - max(f.age)""",
+          ResultAssertions((r) => {
+            r.toList should equal(List(Map("n.age" -> 13L, "n.age - max(f.age)" -> -31)))
+          })) {
+          resultTable()
+        }
+        p("Using the property `n.age`, which is not a grouping key will throw an exception.")
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age - max(f.age)""",
+          ErrorAssertions((t) => {
+            t.getMessage should include("Aggregation column contains implicit grouping expressions.")
+          })) {
+          errorOnlyResultTable()
+        }
+        note {
+          p(
+            """The above query could be rewritten to:
+              #`MATCH (n: Person{name:\"A\"})-[:KNOWS]-(f:Person) WITH n.age AS age, max(f.age) AS oldestFriend RETURN age - oldestFriend`""".stripMargin('#'))
+        }
       }
     }
   }.build()
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
@@ -419,7 +419,7 @@ class AggregatingFunctionsTest extends DocumentingTest {
           })) {
           resultTable()
         }
-        p("Mix of the aggregation function `max` and the literal `1`.")
+        p("Mixing literal `1` with an aggregation function.")
         query("""MATCH (p: Person) RETURN 1 + max(p.age)""",
           ResultAssertions((r) => {
             r.toList.head("1 + max(p.age)") should equal(45L)
@@ -441,8 +441,14 @@ class AggregatingFunctionsTest extends DocumentingTest {
           })) {
           resultTable()
         }
-        p("Using the property `n.age`, which is not a grouping key will throw an exception.")
+        p("Using the property `n.age` will throw an exception, since `n.age` is not a grouping keys.")
         query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age - max(f.age)""",
+          ErrorAssertions((t) => {
+            t.getMessage should include("Aggregation column contains implicit grouping expressions.")
+          })) {
+        }
+        p("`n.age + n.age` is the grouping key, but since the expression is not a literal, parameter, variable or property/map access it can not be used in the expression which contains the aggregation function.")
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age + n.age, n.age + n.age - max(f.age)""",
           ErrorAssertions((t) => {
             t.getMessage should include("Aggregation column contains implicit grouping expressions.")
           })) {
@@ -450,7 +456,7 @@ class AggregatingFunctionsTest extends DocumentingTest {
         note {
           p(
             """The above query could be rewritten to:
-              #`MATCH (n: Person{name:\"A\"})-[:KNOWS]-(f:Person) WITH n.age AS age, max(f.age) AS oldestFriend RETURN age - oldestFriend`""".stripMargin('#'))
+              #`MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) WITH n.age + n.age AS groupingKey, f RETURN groupingKey, groupingKey - max(f.age)`""".stripMargin('#'))
         }
       }
     }
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
@@ -409,12 +409,13 @@ class AggregatingFunctionsTest extends DocumentingTest {
       p(
         """For aggregation expressions to be correctly computable for the buckets formed by the grouping key(s), they have to fulfill some requirements.
           #Specifically, each sub expression in an aggregation expression has to be either:
-          #- an aggregation function, e.g. `SUM(x.a)`,
-          #- a constant, e.g. `0.1`,
-          #- a parameter, e.g. `$param`,
-          #- a grouping key, e.g. the `a` in `RETURN a, count(*)`
-          #- a local variable, e.g. the `x` in  `count(*) + size([ x IN range(1, 10) | x ])`, or
-          #- a subexpression, all whose operands are operands allowed in an aggregation expression.
+          #
+          #* an aggregation function, e.g. `SUM(x.a)`,
+          #* a constant, e.g. `0.1`,
+          #* a parameter, e.g. `$param`,
+          #* a grouping key, e.g. the `a` in `RETURN a, count(*)`
+          #* a local variable, e.g. the `x` in  `count(*) + size([ x IN range(1, 10) | x ])`, or
+          #* a subexpression, all whose operands are operands allowed in an aggregation expression.
           #""".stripMargin('#'))
 
       section("Examples of aggregation expressions.", "grouping-key-examples") {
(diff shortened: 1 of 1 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -69,6 +69,14 @@ class AggregatingFunctionsTest extends DocumentingTest {
         #The latter, `+count(*)+` is an aggregate expression.
         #The matching paths will be divided into different buckets, depending on the grouping key.
         #The aggregate function will then be run on these buckets, calculating an aggregate value per bucket.""".stripMargin('#'))
+    p("""An aggregation functions input expression can contain non-grouping variables, for example `+sum(1 + n.x)+`.
```
tobias-johansson:
```suggestion
    p("""An aggregation function's input expression can contain non-grouping variables, for example `+sum(1 + n.x)+`.
```

tobias-johansson: or perhaps
```suggestion
    p("""The input expression of an aggregation function can contain non-grouping variables, for example `+sum(1 + n.x)+`.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,71 @@ class AggregatingFunctionsTest extends DocumentingTest {
         resultTable()
       }
     }
+    section("Grouping keys", "grouping-keys") {
+      p("""The expressions which does not contain any aggregation function are the grouping keys of the aggregation.
```
tobias-johansson:
```suggestion
      p("""Expressions which do not contain any aggregation function are the grouping keys of the aggregation.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,71 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+          #* A literal
+          #* A Parameter
+          #* A variable - ONLY IF that variable is also one of:
+          #** A projection expression on its own (e.g. the `n` in `RETURN n AS myNode, n.value + count(*)`)
+          #** A local variable in the expression (e.g the `x` in `RETURN n, n.prop + size([ x IN range(1, 10) | x ])`)
+          #* Property access - ONLY IF that property access is also a projection expression on its own (e.g. the `n.prop` in `RETURN n.prop, n.prop + count(*)`)
+          #* Map access - ONLY IF that map access is also a projection expression on its own (e.g. the `map.prop` in `WITH {prop: 2} AS map RETURN map.prop, map.prop + count(*)`)""".stripMargin(
+          '#'))
+      note {
+        p("The same rules apply to expressions in ORDER BY if there is an aggregation in preceding WITH or RETURN clause.")
```
tobias-johansson: Perhaps:
```suggestion
        p("The same rules apply to expressions in ORDER BY if there is an aggregation in the WITH or RETURN clause that it belongs to.")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,71 @@ class AggregatingFunctionsTest extends DocumentingTest {
         resultTable()
       }
     }
+    section("Grouping keys", "grouping-keys") {
+      p("""The expressions which does not contain any aggregation function are the grouping keys of the aggregation.
+          #If there are no grouping keys, the aggregation will be over all rows.
+          #For expressions that contain an aggregation function, the leaves of the expressions must be one of:
```
tobias-johansson: Is there some way to explain the reasoning here, perhaps by grouping the alternatives like

> For expressions that contain an aggregation function, the leaves of the expressions must each be one of:
> * An aggregation
> * A constant, i.e.
>   * A literal
>   * A Parameter
> * A simple grouping key, i.e.
>   * A variable - if that variable is a projection expression on its own (e.g. the `n` in `RETURN n AS myNode, n.value + count(*)`)
>   * Property access - ONLY IF that property access is also a projection expression on its own (e.g. the `n.prop` in `RETURN n.prop, n.prop + count(*)`)
>   * Map access - ONLY IF that map access is also a projection expression on its own (e.g. the `map.prop` in `WITH {prop: 2} AS map RETURN map.prop, map.prop + count(*)`)
> * A local variable, i.e.
>   * A variable introduced in the expression (e.g the `x` in `RETURN n, n.prop + size([ x IN range(1, 10) | x ])`)

Would this help users grasp the reasoning better? I'm not sure, I'll leave it up to you :smile:

tobias-johansson (changes requested):
Looks good! I've made a few suggestions on wording in places

LinneaAndersson:
[Document ambiguous aggregation expressions](https://trello.com/c/MRtx1H10/1181-document-ambiguous-aggregation-expressions)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
LinneaAndersson: Could we link to this document: https://opencypher.org/articles/2017/07/27/ocig1-aggregations-article/ 😅  It's very complex and would be to hard to explain here I think. 

I was thinking that we maybe shouldn't put it in the documentation, but then it is impossible to know when you can use it and when you can't. This at least gives an explanation to people who is _really_ interested in how it's working (if they understand it, hehe.)

Do you have any opinion about this @hvub ?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -69,6 +69,14 @@ class AggregatingFunctionsTest extends DocumentingTest {
         #The latter, `+count(*)+` is an aggregate expression.
         #The matching paths will be divided into different buckets, depending on the grouping key.
         #The aggregate function will then be run on these buckets, calculating an aggregate value per bucket.""".stripMargin('#'))
+    p("""The input expression of an aggregation function can contain non-grouping variables, for example `+sum(1 + n.x)+`.
```
hvub: Does "grouping variable" mean the same as "grouping key"? Or not? If yes, then it is better to use the same term consistently. If no, then grouping variable should be explained/introduce somewhere.

hvub: I guess a "non-grouping variable" is a variable that it not a grouping key. Right? It is probably got to say that, e.g. "Every variable that is not a grouping key is a non-grouping variable."

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
hvub: I would not like that document in a manual. The manual describes the state of the product and how to use it. It does not given technical design background. If there is the desire to provide technical design background for the _really_ interested user, there are other formate, e.g. developer blog posts, where that can be done.

BTW: "the leaves of the expressions" feels also too language-nerdy for the manual.

We can provide some reasoning by staging the explanation a bit. Sketch:

=====

Expressions that contain an aggregation function are aggregation expressions. Simple aggregation expression only an aggregation function an its argument. For instance, `SUM(x.a)` is an aggregation expressions that only consists of the aggregation function `SUM( )` with `x.a` as it argument. Aggregation expression are also allowed to be more complex, where the result of one or more aggregation functions are operands to other computation. For instance, `0.1 * (SUM(x.a) / COUNT(x.b))` is an aggregation expressions that contains two aggregation functions, `SUM( )` with `x.a` as it argument and `COUNT( )` with `x.b` as it argument. Both are operand to a numeric division. The result of this numeric division is an operand to a numeric multiplication. The numeric multiplication has the value `0.1` as its second operand.

For aggregation expressions to be correctly computable for of the buckets formed by the grouping key, they have to fulfill some requirements. Specifically, each operand in an aggregation expression has to be either

- an aggregation function, e.g. `SUM(x.a)`,
- a constant, e.g. `0.1`,
- a parameter, e.g. `$param`,
- a grouping key, e.g. ....
- a local variable, e.g. ..., or
- a subexpression, all whose operands are operands allowed in an aggregation expression.

// pro example

// counter example

Grouping keys themselves can be complex expression. For better query readability, Cypher only recognizes operands in aggregation expressions to be grouping key if the grouping key is either

- A variable - ...
- A property access - ...
- A map access - ...

If more complex grouping key are need as operands in aggregation expression, it is always possible to project them in advance with `WITH`.

// example (hopefully implicitly demonstrates, that the extra project steps help query readability.)

=====

Again this a sketch. No ready-to-be-used text snippet. It is meant as a suggest of how the text could be made more digestible in a way that is also suitable for a manual – at least according to my uneducated understanding of manuals. The idea of the sketch is to stage the explanation a bit more and provide some high level reasons without going into details, e.g. "to be correctly computable", "for better query readability". It certainly needs more editorial love and integration with the other bits. The more concrete the examples the better. Feel free to improve in any way you think suitable.

hvub (changes requested):
I would not link to that other material. I would rather improve the text to be more comprehensible to the non-language nerd person. More concrete suggestion in that respect can be found inline. ;-)

LinneaAndersson:
@tobias-johansson @hvub descritpion contains updated screenshots

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+          ResultAssertions((r) => {
+            r.toList should equal(List(Map("n.age" -> 13L, "n.age - max(f.age)" -> -31)))
+          })) {
+          resultTable()
+        }
+
+        p(
+          """
+            #Grouping keys themselves can be complex expressions.
+            #For better query readability, Cypher only recognizes sub-expressions in aggregation expressions to be grouping key if the grouping key is either:
```
hvub: Typo: The grammatical number of "to be grouping key" is not right.
```suggestion
            #For better query readability, Cypher only recognizes sub-expressions in aggregation expressions to be grouping keys if the grouping key is either:
```

Alternatively, one can fully switch to singular: 
"Cypher only recognizes a sub-expression in aggregation expressions as a grouping key if the grouping key is either"

To the side, I seem to like "recognizes ... as" better than "recognizes ... to be". But there are probably people more word-smithy then me to decide about such editorial details.

hvub (approved):
Looks good to me. :-)

I left one editorial comment. 

I wonder — and that is something to raise with the docs teams, I guess — if negative examples (the ones that do not work) can be presented optically different from examples that actually work. If you just scan the manual for examples, you can easy fall in the trap of assuming that all listed examples work. Hence, some simple optical indicator for negative examples, alarming the reader that the example is not the usual kind, would be helpful.  —— This is just something for the suggestion box. May such indicator already exists, I do not know.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
tobias-johansson: Perhaps it should read:
> The input expression of an aggregation function can contain any expression, including expressions that are not grouping keys. 
> However, not all expressions can be _composed_ with aggregation functions.
> The example below will throw an error since we compose `n.x`, which is not a grouping key, with the aggregation expression `+count(*)+`.
> For more information see <<grouping-keys,Grouping keys>>.

I.e. we don't have to be very specific here (the details are explained in the link) but rather just introduce the idea that some expressions don't go together with aggregation functions.

tobias-johansson (approved):
I think this looks great overall!

LinneaAndersson:
@davidoliverSP2 for the failing example, is there any nicer way to display this? As Hannes mention above, it would be nice with some optically difference from the example that works. I tried to search for it, but I couldn't really find anything.

davidoliverSP2:
> @davidoliverSP2 for the failing example, is there any nicer way to display this? As Hannes mention above, it would be nice with some optically difference from the example that works. I tried to search for it, but I couldn't really find anything.

At the moment, no - we don't have a great way to display this. However, when we move to the new static Cypher Manual repository, we'll have greater control to amend how we display certain examples. I'll log this as a use case for this scenario as a card on our wall.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+          #* an aggregation function, e.g. `SUM(x.a)`,
+          #* a constant, e.g. `0.1`,
+          #* a parameter, e.g. `$param`,
+          #* a grouping key, e.g. the `a` in `RETURN a, count(*)`
+          #* a local variable, e.g. the `x` in  `count(*) + size([ x IN range(1, 10) | x ])`, or
+          #* a subexpression, all whose operands are operands allowed in an aggregation expression.
+          #""".stripMargin('#'))
+
+      section("Examples of aggregation expressions.", "grouping-key-examples") {
+        p("Simple aggregation without any grouping keys.")
```
davidoliverSP2:
```suggestion
        p("Simple aggregation without any grouping keys:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        query("""MATCH (p: Person) RETURN max(p.age)""",
+          ResultAssertions((r) => {
+            r.toList.head("max(p.age)") should equal(44L)
+          })) {
+          resultTable()
+        }
+        p("Addition of an aggregation and a constant, without any grouping keys.")
```
davidoliverSP2:
```suggestion
        p("Addition of an aggregation and a constant, without any grouping keys:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        query("""MATCH (p: Person) RETURN max(p.age) + 1""",
+          ResultAssertions((r) => {
+            r.toList.head("max(p.age) + 1") should equal(45L)
+          })) {
+          resultTable()
+        }
+        p("Subtraction of a property access and an aggregation. Note that `n` is a grouping key.")
```
davidoliverSP2:
```suggestion
        p("Subtraction of a property access and an aggregation. Note that `n` is a grouping key:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n, n.age - max(f.age)""",
+          ResultAssertions((r) => {
+            r.toList.size shouldBe 1
+            r.head.get("n.age - max(f.age)") should equal(Some(-31L))
+          })) {
+          resultTable()
+        }
+        p("Subtraction of a property access and an aggregation. Note that `n.age` is a grouping key.")
```
davidoliverSP2:
```suggestion
        p("Subtraction of a property access and an aggregation. Note that `n.age` is a grouping key:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+            #For better query readability, Cypher only recognizes a sub-expression in aggregation expressions as a grouping key if the grouping key is either:
+            #
+            #- A variable - e.g. the `n` in `RETURN n, n.age - max(f.age)`
+            #- A property access - e.g. the `n.age` in `RETURN n.age, n.age - max(f.age)`
+            #- A map access - e.g. the `n.age` in `WITH {age: 34, name:Chris} AS n RETURN n.age, n.age - max(n.age)`
+            #
+            #If more complex grouping keys are needed as operands in aggregation expression, it is always possible to project them in advance with `WITH`.
+            #""".stripMargin('#'))
+
+        p("Using the property `n.age` will throw an exception, since `n.age` is not a grouping keys.")
```
davidoliverSP2:
```suggestion
        p("Using the property `n.age` will throw an exception, since `n.age` is not a grouping keys:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age - max(f.age)""",
+          ErrorAssertions((t) => {
+            t.getMessage should include("Aggregation column contains implicit grouping expressions.")
+          })) {
+        }
+        p("`n.age + n.age` is not a valid grouping key, since the expression is not a variable, property access or map access. " +
+          "It can therefore not be used in the expression which contains the aggregation function.")
```
davidoliverSP2:
```suggestion
          "It can therefore not be used in the expression which contains the aggregation function:")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/AggregatingFunctionsTest.scala
```diff
@@ -386,5 +394,92 @@ class AggregatingFunctionsTest extends DocumentingTest {
...
+        query("""MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) RETURN n.age + n.age, n.age + n.age - max(f.age)""",
+          ErrorAssertions((t) => {
+            t.getMessage should include("Aggregation column contains implicit grouping expressions.")
+          })) {
+        }
+        note {
+          p(
+            """The above query could be rewritten to:
+              #`MATCH (n: Person{name:"A"})-[:KNOWS]-(f:Person) WITH n.age + n.age AS groupingKey, f RETURN groupingKey, groupingKey - max(f.age)`""".stripMargin('#'))
```
davidoliverSP2: This would be better displayed as a Cypher example, rather than in codeblock markup. I'm not sure on how to embed that in the scala code though!

davidoliverSP2 (approved):
Thanks @LinneaAndersson 

Just a few editorial changes.

PR #1586, opened 2022-08-30
Fixed build issues

martin-neotech:


recrwplay (commented):
tests pass and HTML builds with this update - thanks!

PR #1587, opened 2022-08-31
Users should not use predicates on varlength relationships

arnefischereit:
We have documented this in a central place. Therefore, we should make sure that they visit this place and not just look at the example given here.


Dunfluff (approved):
LGTM

AlexicaWright (approved):
LGTM 👍  but the tests are all red. Perhaps a rebase would resolve?

PR #1590, opened 2022-09-05
Document EXISTS and COUNT subqueries

sherfert:
Replaces #1490

## TODO
~~- [ ] Differences between `COUNT` and `EXISTS`~~
~~- [ ] Optimizations made for `COUNT` internally?~~
- [x] `COUNT` as groupping key
- [x] Double check `size((a)-[r]->(b))` has completely been removed from the docs

The TODOs were copied from the previous PR.
* A am unsure if/how to document differences between `COUNT` and `EXISTS` - I guess that was when they supported different subqueries inside?
* For optimizations: 
  * We have no mention of `GetDegree` anywhere, so I am hesitant to document that optimization.
  * We could mention that `COUNT {} > 0` is handled as `EXISTS {}` internally. Should we?

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -877,6 +877,16 @@ Cypher now supports float literals for the values `Infinity` and `NaN`.
 `NaN` defines a quiet not-a-number value and does not throw any exceptions in arithmetic operations.
 Both values are implemented according to the Floating-Point Standard IEEE 754.
 
+a|
+label:functionality[]
+label:added[]
+[source, cypher, role="noheader"]
+----
+COUNT { (n) WHERE n.foo = "bar" }
+----
+a|
+New expression which returns the number of occurrences of a pattern.
+
 |===
 
 [[cypher-deprecations-additions-removals-4.4]]
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -2855,7 +2855,7 @@ label:new[]
 WHERE EXISTS {...}
 ----
 a|
-Existential sub-queries are sub-clauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
+EXISTS sub-queries are sub-clauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
 
 a|
 label:clause[]
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -885,7 +885,7 @@ label:added[]
 COUNT { (n) WHERE n.foo = "bar" }
 ----
 a|
-New expression which returns the number of occurrences of a pattern.
+New expression which returns the number of results of a subquery.
 
 |===
 
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -2865,7 +2865,7 @@ label:new[]
 WHERE EXISTS {...}
 ----
 a|
-EXISTS sub-queries are sub-clauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
+`EXISTS` subqueries are subclauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
 
 a|
 label:clause[]
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ScalarFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ScalarFunctionsTest.scala
@@ -268,8 +268,8 @@ class ScalarFunctionsTest extends DocumentingTest {
     section("size() applied to pattern comprehension", "functions-size-of-pattern-comprehension") {
       p("""This is the same function `size()` as described above, but you pass in a pattern comprehension.
           #The function size will then calculate on a _list_ of paths.""".stripMargin('#'))
-      function("size(pattern expression)",
-        ("pattern expression", "A pattern expression that returns a list."))
+      function("size(pattern comprehension)",
+        ("pattern comprehension", "A pattern comprehension that returns a list."))
       query("""MATCH (a)
               #WHERE a.name = 'Alice'
               #RETURN size([p=(a)-->()-->() | p]) AS fof""".stripMargin('#'),
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
@@ -51,7 +51,7 @@ class SubqueryExpressionsTest extends DocumentingTest {
     section("EXISTS subqueries", "existential-subqueries") {
       p(
         """An EXISTS subquery can be used to find out if a specified pattern exists at least once in the data.
-          #It serves the same purpose as a path pattern but is more powerful because it allows you to use `MATCH` and `WHERE` clauses internally.
+          #It serves the same purpose as a <<filter-on-patterns, path pattern>> but is more powerful because it allows you to use `MATCH` and `WHERE` clauses internally.
           #Moreover, it can appear in any expression position, unlike path patterns.
           #A subquery has a scope, as indicated by the opening and closing braces, `{` and `}`.
           #Any variable that is defined in the outside scope can be referenced inside the subquery's own scope.
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
@@ -67,7 +67,7 @@ class SubqueryExpressionsTest extends DocumentingTest {
           #  MATCH [Pattern]
           #  WHERE [Expression]
           #}""".stripMargin('#'))
-      p("It is worth noting that the `MATCH` keyword can be omitted in subqueries and that the `WHERE` clause is optional.")
+      p("It is worth noting that the `MATCH` keyword can be omitted in such subqueries and that the `WHERE` clause is optional.")
 
       section("Simple EXISTS subquery", "existential-subquery-simple-case") {
         p(
@@ -145,14 +145,14 @@ class SubqueryExpressionsTest extends DocumentingTest {
       }
     }
 
-    section("COUNT subqueries", "count-subqueries") {
-      p("An count subquery can be used to find out how many times a specified pattern appears in the data.")
+    section("`COUNT` subqueries", "count-subqueries") {
+      p("A `COUNT` subquery expression can be used to to count the number of results of the subquery.")
       functionWithCypherStyleFormatting(
         """COUNT {
           #  [Pattern]
           #  WHERE [Expression]
           #}""".stripMargin('#'))
-      p("It is worth noting that the `MATCH` keyword can be omitted in subqueries and that the `WHERE` clause is optional.")
+      p("It is worth noting that the `MATCH` keyword can be omitted in such subqueries and that the `WHERE` clause is optional.")
 
       section("Simple count subquery", "count-subquery-simple-case") {
         p(
@@ -203,11 +203,11 @@ class SubqueryExpressionsTest extends DocumentingTest {
         section("Using count in SET", "count-subqueries-with-set") {
           query(
             """MATCH (person:Person) WHERE person.name ="Andy"
-              #SET person.howManyDogs = COUNT { (person)-[:HAS_DOG]->(:Dog) } + 1
+              #SET person.howManyDogs = COUNT { (person)-[:HAS_DOG]->(:Dog) }
               #RETURN person.howManyDogs as howManyDogs
       """.stripMargin('#'),
             ResultAssertions(r => {
-              r.toList should equal(List(Map("howManyDogs" -> 2)))
+              r.toList should equal(List(Map("howManyDogs" -> 1)))
             })) {
             resultTable()
           }
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
@@ -28,11 +28,9 @@ class SubqueryExpressionsTest extends DocumentingTest {
     doc("Subquery expressions", "cypher-subquery-expressions")
     initQueries(
       """CREATE
-        #(andy:Swedish:Person {name: 'Andy', age: 36, belt: 'white'}),
-        #(timothy:Person {name: 'Timothy', age: 25, address: 'Sweden/Malmo'}),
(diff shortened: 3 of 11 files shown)


cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/ScalarFunctionsTest.scala
```diff
@@ -268,8 +268,8 @@ class ScalarFunctionsTest extends DocumentingTest {
     section("size() applied to pattern comprehension", "functions-size-of-pattern-comprehension") {
       p("""This is the same function `size()` as described above, but you pass in a pattern comprehension.
           #The function size will then calculate on a _list_ of paths.""".stripMargin('#'))
-      function("size(pattern expression)",
-        ("pattern expression", "A pattern expression that returns a list."))
+      function("size(pattern comprehension)",
```
arnefischereit, line 271: :see_no_evil:

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
```diff
@@ -53,5 +53,17 @@ EXISTS {
...
+MATCH (n) WHERE
+
+COUNT {
+  MATCH (n)-[:KNOWS]->(m) WHERE n.age = m.age
+}
+
+= 1
+RETURN n###
+
+Use a COUNT subquery to count the occurrences of the subquery.
```
arnefischereit: I always thought of the whole thing to be the subquery, whereas this sentence suggests that only the inside is the subquery. One could alternatively write about "expression" or "pattern", but those words are also "overloaded" already...

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -140,5 +144,107 @@ class SubqueryExpressionsTest extends DocumentingTest {
         }
       }
     }
+
+    section("COUNT subqueries", "count-subqueries") {
+      p("An count subquery can be used to find out how many times a specified pattern appears in the data.")
```
arnefischereit:
```suggestion
      p("A `COUNT` subquery can be used to find out how many times a specified pattern appears in the data.")
```
Here you call it "pattern".

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -140,5 +144,107 @@ class SubqueryExpressionsTest extends DocumentingTest {
...
+      functionWithCypherStyleFormatting(
+        """COUNT {
+          #  [Pattern]
+          #  WHERE [Expression]
+          #}""".stripMargin('#'))
+      p("It is worth noting that the `MATCH` keyword can be omitted in subqueries and that the `WHERE` clause is optional.")
```
arnefischereit:
```suggestion
      p("It is worth noting that the `MATCH` keyword can be omitted in such subqueries and that the `WHERE` clause is optional.")
```
or maybe 
```
... in `COUNT` subqueries
```
because that is not allowed in "normal" subqueries `CALL { ... }`, is it?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -140,5 +144,107 @@ class SubqueryExpressionsTest extends DocumentingTest {
...
+
+      section("Simple count subquery", "count-subquery-simple-case") {
```
arnefischereit:
```suggestion
      section("Simple `COUNT` subquery", "count-subquery-simple-case") {
```
... there are more cases like this one.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -140,5 +144,107 @@ class SubqueryExpressionsTest extends DocumentingTest {
...
+            ResultAssertions(r => {
+              r.toList should equal(List(Map("howManyDogs" -> 1, "person.name" -> "Andy"), Map("howManyDogs" -> 0, "person.name" -> "Timothy"), Map("howManyDogs" -> 2, "person.name" -> "Peter")))
+            })) {
+            resultTable()
+          }
+        }
+        section("Using count in SET", "count-subqueries-with-set") {
+          query(
+            """MATCH (person:Person) WHERE person.name ="Andy"
+              #SET person.howManyDogs = COUNT { (person)-[:HAS_DOG]->(:Dog) } + 1
```
arnefischereit: This `+ 1` is probably to indicate that one can also use this in algebraic expressions but it might suggest to some readers that `COUNT` would result in a too low value.

arnefischereit (commented):
Looks good. I would like to see some more consistency in using `COUNT` over "count" and the like.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
sherfert: I think I would call `COUNT { MATCH ... }` a subquery expression, while `MATCH ...` is the subquery.
Analogous `CALL { ... }` would be a subquery _clause_.

So, WDYT about `Use a COUNT subquery expression to count the occurrences of the subquery.` ?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
sherfert: I'll remove the `+1`, I think that is less confusing.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
arnefischereit: Works for me but we should then align the first sentence of the section (see next comment).

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
arnefischereit: :+1:

sherfert:
@arnefischereit 
I repurpose this PR to include both EXISTS/COUNT, since it is easier for me than to rebase all the time.

I should have fixed all your comments now.

Could you comment on this?
> The TODOs were copied from the previous PR.
> 
> * A am unsure if/how to document differences between `COUNT` and `EXISTS` - I guess that was when they supported different subqueries inside?
> * For optimizations:
>   
>   * We have no mention of `GetDegree` anywhere, so I am hesitant to document that optimization.
>   * We could mention that `COUNT {} > 0` is handled as `EXISTS {}` internally. Should we?

arnefischereit:
I do agree that it makes more sense to combine the two PRs. I will have a look at the comments separately. Regarding the TODOs:
1. I do agree with you that this point seems less relevant now. One could imagine to make one "subquery expression" documentation and then present `EXISTS` and `COUNT` as two flavours of this in order to stress them being very similar.
2. I think the idea here was to mention that and also `COUNT { .. } == 0` being equivalent to `NOT EXISTS { ... }`. I think we generally never write about these optimisations that we do in the code, so I would not do it here either. The only benefit the user would get from this is that they could write `COUNT { ... } > 0` instead of `EXISTS { ... }`, knowing that it would be rewritten. I see very little value in that.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,250 @@
...
+          resultTable()
+        }
+      }
+    }
+
+    section("`COUNT` subqueries", "count-subqueries") {
+      p("A `COUNT` subquery expression can be used to to count the number of results of the subquery.")
+      functionWithCypherStyleFormatting(
+        """COUNT {
+          #  [Pattern]
```
arnefischereit: As of what you said, this should read:
```suggestion
          # MATCH [Pattern]
```

arnefischereit (approved):
Looks good. Just one small nit-pick.

cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
```diff
@@ -2855,7 +2865,7 @@ label:new[]
 WHERE EXISTS {...}
 ----
 a|
-Existential sub-queries are sub-clauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
+EXISTS sub-queries are sub-clauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
```
lidiazuin:
```suggestion
`EXISTS` subqueries are subclauses used to filter the results of a `MATCH`, `OPTIONAL MATCH`, or `WITH` clause.
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,250 @@
...
+          ResultAssertions(r => {
+            r.toList should equal(List(Map("name" -> "Peter")))
+          })) {
+          resultTable()
+        }
+      }
+      section("`EXISTS` subquery outside of a `WHERE` clause", "existential-subquery-outside-where") {
+        p(
+          """`EXISTS` subquery expressions can appear anywhere that an expression is valid.
+            |Here the result if the subquery can find the given pattern is returned.""".stripMargin)
```
lidiazuin:
```suggestion
            |Here is the result in case the subquery can find the given pattern:""".stripMargin)
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,250 @@
...
+              Map("name" -> "Peter", "hasDog" -> true)
+            ))
+          })) {
+          resultTable()
+        }
+      }
+    }
+
+    section("`COUNT` subqueries", "count-subqueries") {
+      p("A `COUNT` subquery expression can be used to to count the number of results of the subquery.")
```
lidiazuin:
```suggestion
      p("A `COUNT` subquery expression can be used to count the number of results of the subquery.")
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,250 @@
...
+      functionWithCypherStyleFormatting(
+        """COUNT {
+          #  MATCH [Pattern]
+          #  WHERE [Expression]
+          #}""".stripMargin('#'))
+      p("It is worth noting that the `MATCH` keyword can be omitted in such subqueries and that the `WHERE` clause is optional.")
+
+      section("Simple `COUNT` subquery", "count-subquery-simple-case") {
+        p(
+          """Variables introduced by the outside scope can be used in the `COUNT` subquery without importing them
```
lidiazuin: This sentence is a bit confusing. Was it supposed to be "Variables introduced by the outside scope can be used in the `COUNT` subquery without importing them, >>>as it is<<< necessary with `CALL` subqueries"?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,250 @@
...
+            #} = 1
+            #RETURN person.name AS name""".stripMargin('#'),
+          ResultAssertions(r => {
+            r.toList should equal(List(Map("name" -> "Andy")))
+          })) {
+          resultTable()
+        }
+      }
+      section("Using `COUNT` subqueries inside other clauses", "count-subqueries-other-clauses") {
+        p("`COUNT` can be used in any position in a query, with the exception of administration commands, where it is restricted. We provide a few examples below.")
```
lidiazuin:
```suggestion
        p("`COUNT` can be used in any position in a query, with the exception of administration commands, where it is restricted. 
        #See a few examples below:")
```

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
```diff
@@ -0,0 +1,69 @@
...
+###assertion=returns-one
+MATCH (n) WHERE
+
+EXISTS {
+  MATCH (n)-->(m) WHERE n.age = m.age
+}
+
+RETURN n###
+
+Use an `EXISTS` subquery to test for existence.
```
lidiazuin: This sentence is a bit off. What about something like "Use an `EXISTS` subquery to check the existence of the match" or something like this? I don't know if 'match' is the best word here.

lidiazuin (changes requested):
A few editorial notes. Thank you!

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
sherfert: No, that is not how I meant it. But I can see it is not clear.
I suggest "Here the result is a boolean that shows whether the subquery can find the given pattern."

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
sherfert: This sentence should convey 2 things:
* Variables introduced by the outside scope can be used in the `COUNT` subquery without importing them.
* This is different in `CALL` subqueries. There it is necessary to import to import variables introduced by the outside scope.

I am happy to change "as necessary with `CALL` subqueries" to "as it is necessary with `CALL` subqueries", if you think that is better.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
sherfert: What do you think about 

```
Use an `EXISTS` subquery expression to test for existence of a subquery.
```
?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
lidiazuin: Yep! I think your suggestion makes it clearer now. Thanks!

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/SubqueryExpressionsTest.scala
lidiazuin: Right, got it. But if we say "as it is necessary with `CALL` subqueries", we can leave a margin of ambiguity that might make the sentence look like you're saying that it's the same thing rather than the opposite. What about something like this?

Variables introduced by the outside scope can be used in the `COUNT` subquery without importing them, unlike the case for `CALL` subqueries, as they require importing.

cypher/refcard-tests/src/test/scala/org/neo4j/cypher/docgen/refcard/SubqueryExpressionsTest.scala
lidiazuin: Looks good! I would only add "the" before "existence" - test for the existence of a subquery.

lidiazuin (approved):
LGTM! Thanks

PR #1591, opened 2022-09-06
Alternative relationship types syntax deprecated in 3.2 (#1582)

martin-neotech:
`MATCH (n)-[:A|:B|:C {foo: 'bar'}]-() RETURN n` - deprecated in 3.2, removed in 5.0

Replaced by `MATCH (n)-[:A|B|C {foo: 'bar'}]-() RETURN n`


davidoliverSP2 (approved):
👍

PR #1592, opened 2022-09-06
TEXT index does not support range predicate

burqen:
See this update in mono repo https://github.com/neo-technology/neo4j/pull/16847

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/deprecations-additions-and-compatibility.asciidoc
@@ -610,6 +610,8 @@ MATCH (a), (b), shortestPath((a)-[r*1..1]->(b)) RETURN b
 
 === Updated features
 
+
+
 [cols="2", options="header"]
 |===
 | Feature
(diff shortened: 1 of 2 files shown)


nvitucci (approved):

arnefischereit (changes requested):
Sorry for the late day. This must have fallen through the cracks.

Do you think it makes sense to also mention this change in `deprecations-additions-and-compatibility.asciidoc`?

Also, the cypher ref-card talks about text indexes solving range predicates:

> An TEXT index can be automatically used for the equality comparison when comparing to a string. Note that for example toLower(n.name) = "string" does not use an index.
--
> `MATCH (n:Person) WHERE n.name < "Bob"`
 An index can automatically be used for range predicates. Note that a TEXT index is only used if the predicate compares the property with a string.

davidoliverSP2:
> Sorry for the late day. This must have fallen through the cracks.
> 
> Do you think it makes sense to also mention this change in `deprecations-additions-and-compatibility.asciidoc`?
> 
> Also, the cypher ref-card talks about text indexes solving range predicates:
> 
> > An TEXT index can be automatically used for the equality comparison when comparing to a string. Note that for example toLower(n.name) = "string" does not use an index.
> > --
> > `MATCH (n:Person) WHERE n.name < "Bob"`
> > An index can automatically be used for range predicates. Note that a TEXT index is only used if the predicate compares the property with a string.

From the point of view of the Cypher Refcard, there is no action needed (we're not publishing a Cypher Refcard for 5.0) We will need to port this to its replacement though, the Cypher Cheat Sheet.

I think it would make sense for this to be added to the _deprecations-additions-and-compatibility.asciidoc_ file though. @arnefischereit are you able to help with that?

arnefischereit:
@davidoliverSP2 : Do you think, b9e7026 would suffice?

davidoliverSP2:
Re-raised these changes on #1610, against the 5.0 branch.

PR #1620, opened 2022-09-22
TEXT index does not support range predicate (#1610)

davidoliverSP2:


davidoliverSP2:
Closing, as this has been fixed in #1631

PR #1623, opened 2022-09-27
Updating information about auto-parameterization after 5.0 updates

lidiazuin:
The frontend now enforces auto-parameterization even when the query is partially parameterized.

### DIFF
--- a/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
+++ b/cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
@@ -43,7 +43,7 @@ In these examples, parameters are given in JSON; the exact manner in which they 
 [[cypher-parameters-auto-parameterization]]
 == Auto-parameterization
 
-As of 5.0, even when a query is partially parameterized, Cypher will try to infer parameters anyway.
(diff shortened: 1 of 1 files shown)


cypher/cypher-docs/src/docs/dev/syntax/parameters.asciidoc
```diff
@@ -43,17 +43,11 @@ In these examples, parameters are given in JSON; the exact manner in which they
 [[cypher-parameters-auto-parameterization]]
 == Auto-parameterization
 
-When a query does not use parameters, Cypher will try to infer parameters anyway.
+As of 5.0, even when a query is partially parameterized, Cypher will try to infer parameters anyway.
```
davidoliverSP2:
```suggestion
From version 5 onwards, even when a query is partially parameterized, Cypher will try to infer parameters anyway.
```

davidoliverSP2 (approved):
Thanks @lidiazuin 

I think we need to discuss how we refer to versions now that we're moving to a new release schedule. I think we need to be careful when referring to a minor version, perhaps in this usage, just saying "version 5" is best?

cc @AlexicaWright for Style Guide considerations

PR #1621, opened 2022-09-23
Explain scope of WITH and WHERE together

gem-neo4j:
Add an example to explain how WITH and WHERE interact together when it comes to scope visibility

PR #1627, opened 2022-09-29
Explain scope of WITH and WHERE together

gem-neo4j:
cherry picked: #1621


gem-neo4j:
[Visibility of Variables in a WITH clause](https://trello.com/c/OM5E5fDd/91-visibility-of-variables-in-a-with-clause)

PR #1633, opened 2022-10-03
Update explanation for predicate functions when used with null values.

### LINKED ISSUE #12946: The doc about `in list predicates` is not correct
The doc about in list predicates say:

[all()](https://neo4j.com/docs/cypher-manual/4.4/functions/predicate/)
```
The function all() returns true if the predicate holds for all elements in the given list. null is returned if the list is null or all of its elements are null.

Syntax: all(variable IN list WHERE predicate)
```

It says *all() returns null if all of its elements are null*.
But the tested result is not as expect as it says.

### Actual behavior
image
image

gem-neo4j:
Fixes https://github.com/neo4j/neo4j/issues/12946

### DIFF
--- a/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
+++ b/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
@@ -53,7 +53,8 @@ class PredicateFunctionsTest extends DocumentingTest {
     graphViz()
     section("all()", "functions-all") {
       p("""The function `all()` returns `true` if the predicate holds for all elements in the given list.
-          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null` or if the predicate evaluates to `null` for at least one element and
+          #does not evaluate to `false` for any other element.""".stripMargin('#'))
       function("all(variable IN list WHERE predicate)",
         "A Boolean.",
         ("list",
@@ -77,7 +78,8 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("any()", "functions-any") {
       p("""The function `any()` returns `true` if the predicate holds for at least one element in the given list.
-          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null` and no other has evaluated to true.""".stripMargin('#'))
+          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element
+          #and does not evaluate to `true` for any other element.""".stripMargin('#'))
       function("any(variable IN list WHERE predicate)",
         "A Boolean.",
         ("list",
@@ -179,7 +181,8 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("none()", "functions-none") {
       p("""The function `none()` returns `true` if the predicate does _not_ hold for any element in the given list.
-          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null` and no other has evaluated to true.""".stripMargin('#'))
+          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element
+          #and does not evaluate to `true` for any other element.""".stripMargin('#'))
       function("none(variable IN list WHERE predicate)",
         "A Boolean.",
         ("list",
@@ -206,7 +209,8 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("single()", "functions-single") {
       p("""The function `single()` returns `true` if the predicate holds for exactly _one_ of the elements in the given list.
-          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element
+          #and `true` for max one element.""".stripMargin('#'))
       function("single(variable IN list WHERE predicate)",
         "A Boolean.",
         ("list", "An expression that returns a list."),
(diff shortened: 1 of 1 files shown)


gem-neo4j:
[The doc about in list predicates is not correct](https://trello.com/c/LxKIkvNL/314-the-doc-about-in-list-predicates-is-not-correct)

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -53,7 +53,7 @@ class PredicateFunctionsTest extends DocumentingTest {
     graphViz()
     section("all()", "functions-all") {
       p("""The function `all()` returns `true` if the predicate holds for all elements in the given list.
-          #`null` is returned if the list is `null` or all of its elements are `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null`.""".stripMargin('#'))
```
Lojjs: My experiments show

```
1. RETURN all(x IN [null] WHERE x = 1) => null
2. RETURN all(x IN [null, 1] WHERE x = 1) => null
3. RETURN all(x IN [1, null] WHERE x = 1) => null
4. RETURN all(x IN [2,null] WHERE x = 1) => false
5. RETURN all(x IN [null, 2] WHERE x = 1) => null
6. RETURN all(x IN [ 1, null, 2] WHERE x = 1) => null
```

I can see the logic in that 4. would return `false` as regardless of what unknown the `null` stands for, all elements cannot be 1 if there is a 2. But then I wonder why 5. and 6. doesn't give the same result? That feels a bit like a Cypher bug to me... Do you agree?

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -77,7 +77,7 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("any()", "functions-any") {
       p("""The function `any()` returns `true` if the predicate holds for at least one element in the given list.
-          #`null` is returned if the list is `null` or all of its elements are `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null` and no other has evaluated to true.""".stripMargin('#'))
```
Lojjs: My experiments show

```
1. RETURN any(x IN [null] WHERE x = 1) => null
2. RETURN any(x IN [null, 1] WHERE x = 1) => true
3. RETURN any(x IN [1, null] WHERE x = 1) => true
4. RETURN any(x IN [2,null] WHERE x = 1) => null
5. RETURN any(x IN [null, 2] WHERE x = 1) => null
6. RETURN any(x IN [1, null, 2] WHERE x = 1) => true
```

I think that matches with your sentence, so here I just have a suggestion of a better wording:

```suggestion
          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element and does not evaluate to `true` for any other element.""".stripMargin('#'))
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -206,7 +206,7 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("single()", "functions-single") {
       p("""The function `single()` returns `true` if the predicate holds for exactly _one_ of the elements in the given list.
-          #`null` is returned if the list is `null` or all of its elements are `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null`.""".stripMargin('#'))
```
Lojjs: As expected, this is `null` for all the 6 examples from my experiment, but if you have a `null` and more than one element that fulfills the predicate e.g. `RETURN single(x IN [1, null, 1] WHERE x = 1)` that return `false`.

```suggestion
          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element and `true` for max one element.""".stripMargin('#'))
```

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
```diff
@@ -179,7 +179,7 @@ class PredicateFunctionsTest extends DocumentingTest {
     }
     section("none()", "functions-none") {
       p("""The function `none()` returns `true` if the predicate does _not_ hold for any element in the given list.
-          #`null` is returned if the list is `null` or all of its elements are `null`.""".stripMargin('#'))
+          #`null` is returned if the list is `null` or if one of the comparisons evaluates to `null` and no other has evaluated to true.""".stripMargin('#'))
```
Lojjs: My experiments show

```
1. RETURN none(x IN [null] WHERE x = 1) => null
2. RETURN none(x IN [null, 1] WHERE x = 1) => false
3. RETURN none(x IN [1, null] WHERE x = 1) => false
4. RETURN none(x IN [2,null] WHERE x = 1) => null
5. RETURN none(x IN [null, 2] WHERE x = 1) => null
6. RETURN none(x IN [1, null, 2] WHERE x = 1) => false
```

This also seems to match with your sentence, so just same wording suggestion as above

```suggestion
          #`null` is returned if the list is `null`, or if the predicate evaluates to `null` for at least one element and does not evaluate to `true` for any other element.""".stripMargin('#'))
```

Lojjs (changes requested):
I think we didn't discuss this deep enough in 3 minutes 😆 . I did some experiments and gave you suggestions on more accurate wording. Also I think I might have found a bug. The examples are more for your understanding, you don't have to include them in the docs necessarily.

cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/PredicateFunctionsTest.scala
gem-neo4j: Yeah, that sounds like a bug I agree!

Lojjs (approved):

davidoliverSP2:
This has been recreated in the new repository (https://github.com/neo4j/docs-cypher/pull/226)

As we are no longer publishing from this repo, I'll close this PR

PR #1641, opened 2022-10-13
Reword the cypher runtime section and fix the runtime defaults

renetapopova:


cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -64,46 +64,37 @@ In order to use one or more of these options, the query must be prepended with `
...
 
-Interpreted::
-In this runtime, the operators in the execution plan are chained together in a tree, where each non-leaf operator feeds from one or two child operators.
-The tree thus comprises nested iterators, and the records are streamed in a pipelined manner from the top iterator, which pulls from the next iterator and so on.
-In Neo4j 5.0, the community interpreted runtime has been updated to use the slotted runtime, which used to be an enterprise-only feature.
-The slotted runtime works similarly, except for some additional optimizations regarding how the records are streamed through the iterators.
-This results in improvements to both the performance and memory usage of the query.
+Slotted::
+In the _slotted_ runtime, the operators in the execution plan are chained together in a tree, where each non-leaf operator feeds from one or two child operators.
+The tree comprises nested iterators, which stream records in a pipelined manner starting from the top iterator, which pulls from the next iterator, and so on.
```
pontusmelke: Maybe it is just me but maybe we shouldn't use the phrase _stream records in a pipelined manner_, here. Using _pipelined_ in the sentence makes it confusing that it is the other runtime that is called _pipelined runtime_. Maybe it is enough to say, _stream records from the top iterator, which pulls..._?

cypher/cypher-docs/src/docs/dev/query-tuning.asciidoc
```diff
@@ -64,46 +64,37 @@ In order to use one or more of these options, the query must be prepended with `
...
+The slotted runtime covers all operators and queries. +
+In Neo4j 5, it is the default for Community Edition.
 
 [enterprise-edition]#Pipelined#::
-The pipelined runtime was introduced in Neo4j 4.0 as a replacement for the older compiled runtime used in the Neo4j 3.x versions.
-It combines some of the advantages of the compiled runtime in a new architecture that allows for support of a wider range of queries.
-+
-Algorithms are employed to intelligently group the operators in the execution plan in order to generate new combinations and orders of execution which are optimised for performance and memory usage.
-While this should lead to superior performance in most cases, it is still under development and does not support all possible operators or queries (the slotted runtime covers all operators and queries).
+In the _pipelined_ runtime, the operators are grouped in the execution plan to generate new combinations and orders of execution, which are optimized for performance and memory usage. +
```
pontusmelke: maybe we can use pipelines here instead, something like: _operators are grouped into pipelines ..._?

pontusmelke (approved):
Looks good, I only had some comments about the word pipeline which really isn't part of this change, so feel free to ignore if you don't agree.

PR #1645, opened 2022-10-14
Update version in pom

recrwplay:


davidoliverSP2 (approved):
This looks good to me, but is failing tests 👀

recrwplay:
It was failing already, so at least I didn't technically break anything...

And it looks like this is irrelevant now because of https://github.com/neo4j/neo4j-documentation/commit/8fb94d7051c59a2a92bc48418fd95897ca2a5c64, which updates to `5.2.0-SNAPSHOT`.

PR #1651, opened 2022-10-21
Remove exists() from predicate functions in v5

davidoliverSP2:
The `exists()` function appears to be removed in v5, but the Predicate functions page still has it listed.

@sherfert I've tagged you as a reviewer here, as you added the 'EXISTS' content for v5 - please delegate if this needs to be confirmed by someone else.


davidoliverSP2:
[Remove Predicate function/exists from v5 docs](https://trello.com/c/4e2QPuc3/4425-remove-predicate-function-exists-from-v5-docs)

sherfert:
@davidoliverSP2 I don't think the `exists` function has been removed. I can still execute `MATCH (n) WHERE exists( (n)--(:A) ) RETURN n` without problems.

PR #1653, opened 2022-10-31
Upgrade to properly work with neo4j 5.x

klaren:
* Upgrade junit from 4 to 5
* Reformat all code to align with main

Note! The cypher module is disabled to allow everything to proceed. That will need to be addressed in a separate PR.


recrwplay (approved):
Verified by local build.

The cypher-docs will be added again in another PR in order to generate full docs content.

PR #1656, opened 2022-11-02
Add note about the behavior of the id function on composite databases [5.0]

tobias-johansson:


tobias-johansson:
[[Fabric] id of node different if returned from inside or outside of subquery #12940](https://trello.com/c/jCdsomhR/468-fabric-id-of-node-different-if-returned-from-inside-or-outside-of-subquery-12940)

recrwplay:
- https://github.com/neo4j/docs-cypher/pull/162

PR #1664, opened 2022-11-10
fixing deprecated label formatting

lidiazuin:
Same change is also reflected in docs-cypher: https://github.com/neo4j/docs-cypher/pull/176


recrwplay (approved):
Perfect - thanks!

PR #1666, opened 2022-11-15
use longer password than 7 characters in test

HannesSandberg:


mnd999 (approved):
LGTM

lidiazuin:
Thanks @HannesSandberg . This feature is only being released in 5.3, right? So we shouldn't merge this PR now. 
Also, these changes are related to the Cypher Manual, which will be published from the docs-cypher repo from 5.2 on, making Neo4j-documentation deprecated. 
We started doing the same work in this PR https://github.com/neo4j/docs-operations/pull/79, so maybe we just need to update it with the other pages that you covered here?

HannesSandberg:
@lidiazuin 

> This feature is only being released in 5.3, right? So we shouldn't merge this PR now.

True, but we are not breaking anything by increasing the length of the password in previous versions as well.

Sounds good to add it to the other PR. The reason that I created this PR is that I got pinged by Gustav because this build is failing. But maybe that does not matter? https://live.neo4j-build.io/buildConfiguration/Documentation_Neo4jDocumentationTests_Neo4jDevDocsTest/14725847?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

lidiazuin:
> Sounds good to add it to the other PR. The reason that I created this PR is that I got pinged by Gustav because this build is failing. But maybe that does not matter? https://live.neo4j-build.io/buildConfiguration/Documentation_Neo4jDocumentationTests_Neo4jDevDocsTest/14725847?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

So your changes here are supposed to fix this build failure @HannesSandberg ? 

I was just wondering if adding this kind of information in the dev branch, which is currently related to 5.1, would cause a "leak" about stuff that is still coming up? 🤔

HannesSandberg:
yes, thats the plan. Yes, it might leak some information that we are expecting longer passwords. Not sure what our policy is about things like this.

mnd999:
This whole repo is public, so I think that ship has sailed.

lidiazuin:
I asked Luisa about that, but if this is the case, then let's enjoy the ride (or sail) 😅

lidiazuin (approved):
LGTM. We just need to make sure this is also documented here https://github.com/neo4j/docs-operations/pull/79

